PR #280, opened 2015-02-02
Adding support for source code formatting using Scalariform

newca12:
This should have been done before.
Automatic code formatting is step 0 of code quality.
Rules used here are the one used by typesafe + DoubleIndentClassDeclaration


shetzl:
In principle I am in favour of automatic code formatting, I have found incoherent formatting a little annoying on a number of occasions when reading the code (though not annoying enough to correct it manually ;-).

Concerning the actual setup I have one worry:
1) If I understand correctly the setup you suggest has the effect of automatic reformatting whenever sbt compile is called. I usually use vim for code editing and I have many source files open at the same time when I do "sbt compile". Automatic reformatting will then result in numerous "do you want to reload?"-messages from these vim-instances. This would be very unpractical. AFAIK, Giselle (and maybe others?) use the same setup. I would suggest to configure sbt-scalariform so that it must be called explicitely.

I would like to raise the following additional points:
2) There does not seem to be a limit to the number of characters in a line, for example CutIntroduction.scala:209 is very long now. Is it possible to set such a limit, e.g. to around 100?
3) I personally have the impression that the replacement of => by the unicode double arrow does not result in an increase of readability. This character looks quite small in my browser, I had to look twice to recognize it.
4) I have always found that a space after "(" and before ")" increases legibility.

As always when it comes to style, opinions depend a lot on personal taste. If there are others with strong and differing opinions about 2-4 I am happy to succumb to them.

Everybody, please have a look at your favourite part of gapt as reformatted in this pull request and tell us what you think.

quicquid:
Hi Olivier & Stefan!

I just had a look at my code, it looks quite nice. I have to agree
though, that the unicode => feels a bit strange to me, especially
because during development, I would have a mix of unicode and ascii
notation. Therefore i'd prefer the setting

".setPreference(RewriteArrowSymbols, false)"

in the config file.

If you are developing with emacs, you have the same reload problem as in
vi, when the reformatting is part of the build process. What about
making it its own sbt target? Is it possible to have a hook on git push
to warn you when the code needs reformatting? Then we could ensure that
pushes are always properly formatted.

cheers, Martin

gisellemnr:
Hello all,

I agree that gapt needs to have a more uniform formatting, but I am not
sure if doing this automatic is the way to go.

I completely endorse Stefan's point #1. Having the files edited by the
compilation process will be very annoying.
I also agree with #2 and #3: the long lines do not contribute to the
readability of the code at all [1], and I wouldn't like to mix unicode and
ascii characters, specially because I cannot type this character while
coding, which means my files will be *always* changed during compilation.
Regarding #4, I usually use spaces before "(" and after ")", but this is
only cosmetics, as long as the spaces are there ;-)

My suggestion would be to have code guidelines for gapt and to have one
initial commit changing all the files to use these guidelines [2] (also
because commits involving changes in the code itself *and* the coding style
are harder to understand). I believe once this is settled, keeping the
guidelines is a no-brainer, as we can just look around in the code to see
how things are formatted. Each person can also configure their editor of
choice to check for these formats if they wish. And whenever a commit that
has not the correct format is pushed, the person that identifies it can
comment and ask the author to fix it.
We are currently few people and I think it's possible to have such an
agreement. For outside contributions, their pull requests have to be
checked by us anyway before pushing, so we can also check for the
formatting and indicate our rules (which should be documented somewhere) to
the contributor.

What do you think?

Best,
Giselle

[1] I am the author of the line Stefan mentioned, but since I never saw how
to deal with many arguments in gapt, I just left it in one line like all
the other methods, but I would be glad to transform into something with
multiple lines: one for each parameter, all aligned with the ( of the args
list, for instance.
[2] The initial commit could be the one from Olivier, given a few changes,
but without adding the scalariform to sbt.

frehn:
I agree with running automatic code formatting tools manually and checking in prettier source code. I am against running it automatically for the reasons that have been given already, and the additional reason that I think it is sometimes (maybe even often?) the case that the "correct" formatting for some small part of code is decided ad-hoc by the programmer, and will be done wrongly when subjected to automatic code formatting (see the "long-line-problem" raised above).

A question regarding this pull request: The GAPT wiki contains a section on code conventions [1]. How does the pull request relate to these conventions?


OT: I believe, just as checkin comments and submitted issues, some guidelines should be followed when submitting a pull request. I think the submission should contain a concise description (i.e. an "abstract") of the changes, in particular relating to existing aspects of GAPT that are affected by the pull request. It should *not* be necessary to read the whole diff to see what the main changes are.

In this case, I would have preferred a submission along the lines of
"This patch adds automatic code formatting to GAPT via scalariform (link to scalariform homepage). Scalariform is configured to run at compile time via sbt. The patch contains not only the addition of scalariform to the sbt build process, but the GAPT code after a run of scalariform as well. The included scalariform configuration is such that it enforces "Unified indentation" from the GAPT code conventions at [1], but does not enforce the other points of the GAPT code conventions since this is not supported by scalariform."

Also, to be quite honest, I find comments such as "This should have been done before" rather offensive, and think that they have no place in the description of a pull request.

[1] https://github.com/gapt/gapt/wiki/ForDevelopers#code-conventions

newca12:
@Daniel "This should have been done before" mean, "sorry this is a huge change in the code base because this should have been done before". No offense intended.
I agree regarding your other comments about the formalism of the pull request submission.
 
* Add a max line length preference is a long time issue  : https://github.com/mdr/scalariform/issues/68
* OK to remove unicode =>
* Automatic formatting on compilation should not be an issue, scalariform modify files only and only if there is something to do. You should not have this "reload message" all the time. In practice you should not have "reload message'" at all.

You could also look at scalariform doc to help your editor (vi, emacs, etc) use scalariform format but this is not mandatory.

We are a huge team with also vim and emacs users, I have never heard anyone complain about automatic formatting on compilation.
However if this still really an issue for you I am of course very open to the propositions you made
* remove automatic formatting on compilation
* initial re-formating commit or a git hook (the later is of course a better option)
* add a dedicated sbt target for scalariform

But suppose we have this hook and let developpers working with their own format.
How do you make a clean git diff without annoying formatting issues ?

frehn:
@Olivier: Thanks for the clarification, and sorry for misunderstanding and taking offense - I was certainly too thin-skinned there.

One approach would be to try the "on build scalariform", as proposed in the pull request, for a while (say a month) and then discuss whether to keep it like that or to use another one of the suggested methods.

Another question relating to my earlier comment: One potential problem I see is that if there are some lines of code that I want to format in a certain way, but scalariform disagrees with my formatting. It seems that there is no reasonable way to pursue "my formatting", since even if I check in the source code in "my formatting", someone else will re-format it with scalariform and destroy "my formatting". Olivier, do you know how this kind of problem is dealt with? Or is there the assumption that this kind of situation never occurs (since e.g. scalariform always formats code "good enough")?

newca12:
You can do that with comments
// format: OFF    <-- this directive disables formatting from this point
// format: ON   <-- formatter resumes from this point

frehn:
Thanks, Olivier. That certainly settles my doubt - I'd personally be fine with using scalariform, then!

gisellemnr:
I would still vote against using an automatic formatting tool... Adding
comments to the source only to have the tool not formatting this part of
the code looks like a bad hack, apart from adding ad-hoc information that
does not belong there. I believe also that the reloading of files will
still occur many times, as we need very often to debug and no one usually
pays attention to formatting then, only when actually committing changes.

If someone thinks he/she is not disciplined enough to keep the coding
style, we can have a dedicated sbt target for that. But I guess there will
be some point in which looking at the screen beats an automated tool (such
as in the case of the long lines for example).

gebner:
-1 from me.  While I agree in general that consistent formatting is nice, I think that this is neither the right workflow, nor an accepted formatting style.

As for automatic reformatting on compile, I can't imagine how this works well with any editor.  Reformatting on commit seems to be a more traditional and safe solution -- we could even make Jenkins fail the build if the formatting is off ([just like the play framework does](https://github.com/playframework/playframework/blob/9b33c199a7ac942e9d7ce63d1f20eb0568eefaa5/framework/validatePullRequest)).

I just tried editing a file, compiling, then editing it again, and vim only gives this warning after I tried to save it the second time--I don't think this is very comfortable:
```WARNING: The file has been changed since reading it!!!```

On the other hand, an astonishing number of scala projects seem to run scalariform on compile, so there might be some vim setup to make this work.

>  Rules used here are the one used by typesafe + DoubleIndentClassDeclaration

Source?  I checked activator and playframework, and they definitely don't use this style...

I'd recommend following the [Scala Style Guide](http://docs.scala-lang.org/style/) instead, i.e. [scalariform default settings + DoubleIndentClassDeclaration](http://mdr.github.io/scalariform/#id1)--i.e. no unicode symbols and no parameter/case alignment.

BTW, running sbt compile on your branch still reformats 3 files.

newca12:
Source : akka project
https://github.com/akka/akka/blob/master/project/Formatting.scala

shetzl:
My understanding of these //format: OFF and //format: ON lines is that they are to be used only as an emergency measure. Consider for example the code used for illustrating this feature at http://mdr.github.io/scalariform/. It is certainly not intended to put these comments before and after every long line.

I think there are two different questions:
1) Do we want to use scalariform to enforce a uniform code formatting?, and
2) If yes, how do we want to use it?

I think that we agree on 2) in so far that - if we use it - it should not be run on "sbt compile" but instead via a designated sbt target. Our convention should then be that every commit should run sucessfully not only trough "sbt compile" and "sbt test" but also trough "sbt scalariform" (or however this target will be named). At some later point, and indepedently of that, we may want to enforce this (via Jenkins or commit hooks).

Concerning 1) I think there is a good case to make in favour of scalariform. It helps code readability and leads to a significantly more polished appearence of the code. The disadvantage is that scalariform may sometimes mess up formatting chosen judiciously by the programmer. I personally can live with the long lines. If nothing worse than that happens I think the benefits of using scalariform largely outweigh its disadvantage.

Giselle, I am not sure what you meant by being "against using an automatic formatting tool", is that opposition to 1) or to running scalariform on sbt compile?

gisellemnr:
In principle it's an opposition to 1.
If we use scalariform as an sbt target, what would happen with commits that
do not run successfully through it? I can see a few scenarios:

1. Jenkin fails: this is an overkill for only formatting problems in my
opinion.
2. commit/push fails (is this possible?): this might become annoying as the
developers will have to waste some time changing whatever formatting chosen
by something that would pass scalariform.

IMHO the ideal solution would be to have the sbt target for scalariform for
the moments we want to check if the code is fine, and decide if the
warnings/errors are worth keeping it. But I wouldn't like to have it
running automatically at any moment because I am afraid we will disagree
with it every now and then.

shetzl:
My impression is that the changes introduced by scalariform to the code are quite mild anyway. I tried to find a point where I would disagree with the formatting but the worst I could come up with in the above diff was that it changed your manually edited CutIntroduction:204-205 to the single line CutIntroduction:209. As said earlier, if that's the worst, I'd be happy to live with it.

Can you find any formatting introduced in the above diff with which you strongly disagree and which is not sufficiently rare so that we would want to put //format:OFF and //format:ON around it?

gisellemnr:
It's hard to see, since the commit changes many lines in each file... but
apart from the long lines and some spaces removed before "(", it looks ok.
What would be the policy then? When a commit fails on the sbt scalariform
target, what happens?

shetzl:
The policy is that only such code which is correctly formatted (i.e. where sbt scalariform runs through without making any changes) should be committed. For the time being I would suggest that every comitter is responsible on his/her own for fulfilling this. Just run "sbt scalariform" in addition to "sbt compile" and "sbt test" before each commit.

I would expect this to work sufficiently well. In the case this is forgotten by developer A (which should not happen anyway), then the next developer B will notice this when running "sbt scalariform". With git's flexibility it should be no problem to separate the changes thus created in B's working directory into two commits: one for reformatting A's changes and one for the actual changes of B (if he/she has already made some).

Having Jenkins fail on that seems a bit harsh to me too. In principle it should be possible to install a pre-commit hook that runs "sbt scalariform", "sbt compile" and "sbt test". However, having this done automatically may be a disadvantage in certain situations (e.g. when working on the user guide).

gisellemnr:
I am ok with that... if there was a way to turn off the thing that makes
the lines long, it would be better. But if not, as you said, it's a small
price to pay.

shetzl:
Ok, so after the discussion here and with Gabriel in person I think we have come to a compromise that will work for all:

1) no invocation at sbt compile nor sbt test. Instead we should have a dedicated sbt-target for invoking scalariform.
2) set RewriteArrowSymbols to false
3) set spaceInsideParentheses to true

If possible, split the above commit into two: 1. adding sbt-scalariform support and 2. reformatting the entire code.

Olivier, if you can modify the pull request accordingly, I'd be happy to merge it.

newca12:
I am delighted that consensus has emerged. I will modify the pull request accordingly.
Thank you.

PR #286, opened 2015-02-07
Fix issues when external binaries are not found

newca12:
Using public continuous integration service has revealed two leftover requirements on external binaries.


newca12:
Thanks Stephan for the merge.

PR #288, opened 2015-02-09
Fix deprecations

newca12:
This remove all deprecations in gapt (excepted runOnProver9Proofs which would better be rewritten)
Note the use of Duration instead of Long/Int as it is the natural type used by Await.result and the natural type to use for duration anyway.

### DIFF
--- a/src/main/scala/at/logic/utils/executionModels/timeout.scala
+++ b/src/main/scala/at/logic/utils/executionModels/timeout.scala
@@ -1,8 +1,11 @@
 package at.logic.utils.executionModels
 
-package timeout {
+import scala.concurrent._
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import ExecutionContext.Implicits.global
 
-  class TimeOutException extends Exception
+package timeout {
 
   /**
    * runs f with timeout to
@@ -20,29 +23,8 @@ package timeout {
    * }
    */
   object withTimeout {
-    def apply[T]( to: Long )( f: => T ): T = {
-      var result: Either[Throwable, T] = Left( new TimeOutException() )
-
-      val r = new Runnable {
-        def run() {
-          try {
-            result = Right( f )
-          } catch {
-            case e: Exception =>
-              result = Left( e )
-          }
-        }
-      }
-
-      val t = new Thread( r )
-      t.start()
-      t.join( to )
-      if ( t.isAlive() ) {
-        t.stop()
-      }
-
-      if ( result.isLeft ) throw result.left.get
-      else result.right.get
+    def apply[T]( to: Duration )( f: => T ): T = {
+      Await.result( Future { f }, to )
     }
   }
 
--- a/src/main/scala/at/logic/algorithms/cutIntroduction/CutIntroduction.scala
+++ b/src/main/scala/at/logic/algorithms/cutIntroduction/CutIntroduction.scala
@@ -20,6 +20,8 @@ import at.logic.provers.maxsat.MaxSATSolver
 import at.logic.provers.minisat.MiniSATProver
 import at.logic.transformations.herbrandExtraction.extractExpansionSequent
 import at.logic.utils.executionModels.timeout._
+import java.util.concurrent.TimeoutException
+import scala.concurrent.duration._
 
 class CutIntroException( msg: String ) extends Exception( msg )
 class CutIntroUncompressibleException( msg: String ) extends CutIntroException( msg )
@@ -46,7 +48,7 @@ object CutIntroduction {
    * @throws CutIntroException when the proof is not found.
    */
   def one_cut_one_quantifier( proof: LKProof, verbose: Boolean ) =
-    execute( proof, false, 3600, verbose ) match {
+    execute( proof, false, 1 hour, verbose ) match {
       case ( Some( p ), _, _, None ) => p
       case ( None, _, _, Some( e ) ) => throw e
       case _                         => throw new CutIntroException( "Wrong return value in cut introduction." )
@@ -64,7 +66,7 @@ object CutIntroduction {
    * @throws CutIntroException when the proof is not found.
    */
   def one_cut_one_quantifier( es: ExpansionSequent, hasEquality: Boolean, verbose: Boolean ) =
-    execute( es, hasEquality, false, -1, 3600, verbose ) match {
+    execute( es, hasEquality, false, -1, 1 hour, verbose ) match {
       case ( Some( p ), _, _, None ) => p
       case ( None, _, _, Some( e ) ) => throw e
       case _                         => throw new CutIntroException( "Wrong return value in cut introduction." )
@@ -79,7 +81,7 @@ object CutIntroduction {
    * @throws CutIntroException when the proof is not found.
    */
   def one_cut_many_quantifiers( proof: LKProof, verbose: Boolean ) =
-    execute( proof, true, 3600, verbose ) match {
+    execute( proof, true, 1 hour, verbose ) match {
       case ( Some( p ), _, _, None ) => p
       case ( None, _, _, Some( e ) ) => throw e
       case _                         => throw new CutIntroException( "Wrong return value in cut introduction." )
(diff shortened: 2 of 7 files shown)


frehn:
Thank you Olivier!

I compiled on my machine and verify that this fixes all deprecation warnings.

Olivier, I did notice that your patch introduces some feature warnings like this:

[warn] CutIntroduction.scala:69: postfix operator hour should be enabled
[warn] by making the implicit value scala.language.postfixOps visible.

I suppose we should fix these warnings as well (right?), either by using non-postfix notation, or by enabling postfix notation explicitely as suggested by the warning.

I myself have no opinion on which option to take. Everyone, please feel free to comment.

newca12:
You're right, there is already a bunch a features warnings but this is not a reason to add more.
We need to choose
- Force the non post-fix annotation, add the import or compile with -language:postfixOps
idem for reflectiveCalls, implicitConversions and existentials.
Using explicit import seem the the most consensual approach for the last three.
Using postfix operator or not is really a matter of taste.
Maybe you can merge this pull request and we will fix all features warnings in a future separate pull request when we all agree.

src/main/scala/at/logic/utils/executionModels/timeout.scala
```diff
@@ -20,29 +23,8 @@ package timeout {
...
-      t.start()
-      t.join( to )
-      if ( t.isAlive() ) {
-        t.stop()
-      }
-
-      if ( result.isLeft ) throw result.left.get
-      else result.right.get
+    def apply[T]( to: Duration )( f: => T ): T = {
+      Await.result( Future { f }, to )
```
gebner: This is very nice.  Does it actually kill the thread computing `f`?  I can't figure that out from the API docs for Await.result.

Right now, [we expect that the killed thread receives a ThreadDeath exception](https://github.com/gapt/gapt/blob/17deff3c35ac867bade52b7cf4a5c9b83d65cdd5/src/main/scala/at/logic/provers/maxsat/MaxSAT.scala#L333) which allows us to kill spawned prover9 processes.

frehn:
I am fine with Olivier's suggestion of merging this pull request and dealing with the feature warnings later. It would be good if someone can still check whether the Await.result behaviour is as intended, to answer Gabriel's question (maybe by constructing a toy example in the CLI?)

newca12:
@gebner good catch, I will check this
@frehn I will adapt this pull request if necessary or let you know if it is ok as is

frehn:
FYI, the code

import scala.concurrent.duration._
import scala.concurrent._
import scala.concurrent.ExecutionContext.Implicits.global

def test() : Int = {
  try {
    while (true) {}
  } catch {
    case tde: ThreadDeath => println("DEATH1")
    case _ => println("SOME EXCEPTION1")
  }
  println("EXITING")
  0
}
try {
  Await.result( Future { test }, 5 seconds )
} catch {
    case tde: ThreadDeath => println("DEATH2")
    case _ => println("SOME EXCEPTION2")
}

yields "SOME EXCEPTION2" after 5 seconds.

newca12:
@frehn Thanks, indeed the thread computing f will not be killed after the timeout.
This remember me this old issue : https://github.com/Paradoxika/Skeptik/issues/62
My conclusion is the same. I will try to resume the situation.
Thread.stop and ThreadDeath are very convenient but also inherently unsafe.
[Thread.stop() method has been deprecated for 10 years now](http://docs.oracle.com/javase/1.5.0/docs/guide/misc/threadPrimitiveDeprecation.html).
[In Java 8 there were plans to "de-implement it"](http://cs.oswego.edu/pipermail/concurrency-interest/2013-December/012028.html) but in Java 8 Update 31 it's still there.
**It is not longer possible** to interrupt a thread with blocking code and expect doing something like sanitization by just catching a ThreadDeath like exception.
The only way to imitate the actual behavior is : 
1. Use an interruptable Future like the [one](https://github.com/Paradoxika/Skeptik/issues/62) by Victor Klang
2. **that is the most important** use interruptable function for long computation in gapt. The code must be interruptable mean that in some strategic points we must be able to check a variable in order to determine that we must stop the computation ASAP and eventually make some cleanup.

Remove all deprecations in gapt will need more work than expected.
I will update this pull request later this week.

gebner:
> In Java 8 there were plans to "de-implement it"

This is a somewhat misleading mailing list post; he mistook the removal of Thread.stop(Throwable) for the removal of Thread.stop()--read the follow-ups.  I don't think Thread.stop() is going away soon.

> 2. that is the most important use interruptable function for long computation in gapt.

I don't think this is a very practical solution for gapt.  We don't really have any main loop with many short iterations, where we could just check a boolean flag after every iteration.  It's not at all clear to me where to put the checks (and where to put the flag).

I guess the only "clean" way of doing timeouts in testing is to run each test in a separate VM, and then killing the VM if it takes too long.

frehn:
Maybe it is possible, in the case of terminating a prover9 process, to solve the problem without interrupting a thread? I am thinking of the following solution.

Where we currently run 

Await.result( Future { f }, to )

with a function f, we could run

val o = some object
Await.result( Future{ o.run() }, to )
o.timeoutHappened()

Where o.run() starts a prover9 process, and loops/calls Thread.sleep() until either the process is finished, or timeoutHappened() is called (this presumably sets some mutexed flag in o).  In the latter case, o.run() kills the prover9 process.

Would that work? Would it be clean?

quicquid:
Hi Daniel,

External provers usually have a timeout value (e.g. prover9's -t) which
could be made visible in the Prover interface. Generally, a bash script
performing the kill for external utiities could be added[1]. Where I
don't have a proposition for is our own code like Robinson2LK, which
have high resource requirements.

cheers, Martin

[1] perhaps something like discussed here?
http://stackoverflow.com/questions/5161193/bash-script-that-kills-a-child-process-after-a-given-timeout

shetzl:
I have pushed the two uncontroversial commits to master.

Concerning the use of Thread.stop, note that our use case is the following: we want to run arbitrary code for at most a specified amount of time and if a timeout happens kill that computation (plus all resources it has allocated). As this should work for arbitrary code modifying it to check a flag at regular intervals is not an option.

We are all aware of this not being clean concurrent programming but it is not intended as such, it is a mechanism for carrying out computational experiments - a user will hardly ever use it. We use this for experiments with gapt on a high number of examples (hence the importance of freeing the resources again) and on algorithms whose performance is difficult to predict (hence the timeout).

If there is a way to realize this in a non-deprecated way: good! If not, I would suggest to stay with Thread.stop. I would expect this use case to be sufficiently frequent so that, in the latter case, Thread.stop will not be removed.

newca12:
Sorry for the delay ...
There is no way to stop a block of arbitrary code in a non-deprecated way thus I close this pull request.
I'm reluctant to call "arbitrary code" the gapt code, arbitrary code could be non interruptible, good code shouldn't be non interruptible, gapt code shouldn't be non interruptible.
I wish to emphasise that stopping external program is not a problem at all, the scala.sys.process.Process is well coded and interruptible so Thread.stop is necessary only for long running and non interruptible computation in gapt code.
I'm a little bit disapointed by your comments thus to close the subject, can someone show me an algorithm in gapt that cannot be implemented in a non interruptible way ? 
Can we agree that ceteris paribus an interruptible implementation of an algorithm is better than a non interruptible implementation ?

gebner:
> Can we agree that ceteris paribus an interruptible implementation of an algorithm is better than a non interruptible implementation ?

I strongly disagree.  As far as I can tell, most of our runtime is spent in various recursive functions -- hence we would need to make every function interruptible.  How would we go about doing this?  We could add a dynamic variable (a thread-local global variable) that holds an interrupt flag; and then we would need to check this new variable in *every single function*.

I don't think this is worth the implementation effort, the runtime performance loss, and the maintainance overhead.  Quite frankly, we're not in the business of writing *interruptible* proof transformations, simply writing *reliable* ones is already hard enough.

Also fundamentally, I think this is the wrong layer to implement interruptibility in.  The nice thing about functional programming (and with some unfortunate exceptions, we're mostly following that paradigm) is that we can give our code a nice denotational semantics, where we can interpret each scala function as an actual function, and then *reason* about the code this way.  This seems to fundamentally at odds with introducing a non-deterministic interrupt flag.  (That's why I think this should be implemented at a layer which deals with explicitly imperative code, i.e. at the VM or operating system level.)

PR #299, opened 2015-03-03
Format also test files

newca12:
Did you noticed we actually don't format test files.
The task scalariformFormat don't format test files.
The task test:scalariformFormat need to be used to format test files.

If you consider running these two tasks is cumbersome we probably can make a custom sbt task to join them.
(FYI the defaultScalariformSettings that we disabled to avoid editor reloads run both tasks automatically)

### DIFF
--- a/scalariform.sbt
+++ b/scalariform.sbt
@@ -1,5 +1,7 @@
 import scalariform.formatter.preferences._
 
+addCommandAlias("format", "; scalariformFormat ; test:scalariformFormat")
+
 defaultScalariformSettings
 
 ScalariformKeys.preferences := FormattingPreferences()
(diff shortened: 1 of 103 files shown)


shetzl:
Dear Olivier,

We discussed this in our meeting today and we all think that also the test code should be formatted. In addition, we all would prefer a single sbt target for formatting both main and test code. If you modify the pull request accordingly, I'd be happy to merge it.

Thanks!

newca12:
Thanks Stefan, 
I have adapted this pull request accordingly.
As an alias the name "format" has been chosen intentionally stupid simple.
Therefore the command to use before commit is now simply : sbt format

PR #302, opened 2015-03-09
Sat4j

newca12:
This pull request add a non intrusive fallback to minisat with an embedded Sat4j.
A maybe too verbose warning will tell you each time the Sat4j fallback is used.
The objective is to make minisat installation not mandatory.
If this pull request is accepted, the documentation should probably be modified accordingly.


frehn:
The refactor can be found at commit 1579358ad2c635835eb60de93cb5a8dd5763ff2b; please have a look at it.

newca12:
I have updated the pull request accordingly using a factory.

PR #303, opened 2015-03-14
Ivytorobinson

newca12:
The issue with the junitxml reporter that motive the rollback from specs2 3.0 to specs2 2.4.17 seem to be fixed in 3.0.1
We can now use the new Fragments.foreach feature of specs2 3.x to remove lot of duplicate code in gapt tests.
This pull request show how this can be done using IvyToRobinsonTest as an example.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -102,10 +102,10 @@ lazy val releaseDist = TaskKey[File]("release-dist", "Creates the release tar ba
 
 lazy val testDependencies = Seq(
   "junit" % "junit" % "4.12",
-  "org.specs2" %% "specs2-core" % "2.4.17",
-  "org.specs2" %% "specs2-matcher" % "2.4.17",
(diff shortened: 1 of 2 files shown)


shetzl:
Looks good to me. Just out of curiousity: what is the difference between using Fragments and using a loop as e.g. in testing/src/test/scala/TstpProver9Import.scala ?

gebner:
Mmmh, apparently you can no longer do it like in TstpProver9Import, you have to use Fragments.foreach now.

This PR breaks the testing code (apparently travis doesn't even try to compile it, I'll try to fix that):
```
> testing/test:compile
[info] Compiling 3 Scala sources to /home/gebner/gapt/testing/target/scala-2.11/test-classes...
....
[error] 8 errors found
[error] (testing/test:compile) Compilation failed
[error] Total time: 2 s, completed Mar 16, 2015 2:27:59 PM
```

PR #324, opened 2015-05-01
Reduced redundant binary detection

newca12:
Using a val instead of a def for isInstalled is more appropriate.


shetzl:
Merged into master after code reformatting and solving conflict with recent commit 133312b.

Issue #315, opened 2015-03-30
Weird top/bottom handling in prover9


gebner:
See Prover9TermParser.scala:

```
  //we don't have top and bottom in the algorithms, so we simulate it
  val topformula = { fol.Or( fol.TopC, fol.Neg( fol.TopC ) ) }
  val bottomformula = { fol.And( fol.BottomC, fol.Neg( fol.BottomC ) ) }
```

frehn:
Thanks, Gabriel. I changed this in d3d74d7492e8d3701007d0c320d649a90ab86b51 - let's see whether the next round of regression tests will find some problems with this.

gebner:
:+1: Let's hope for the best!

quicquid:
I think it will work - the algorithms i had in mind when i wrote this were cnf / pcnf, which are fixed now.

Issue #309, opened 2015-03-20
SAT solvers should interpret BottomC and TopC as false and true


gebner:
This should say it all:
```
class SatTest extends Specification {
  "sat4j" should {
    "say bottom is invalid" in {
      new Sat4j().solve(BottomC) must beNone
    }
  }

  "minisat" should {
    "say bottom is invalid as well" in {
      new MiniSAT().solve(BottomC) must beNone
    }
  }
}
```

Kind of unexpected result:
```
[info] SatTest
[info] 
[info] sat4j should
[error]  x say bottom is invalid
[error]   'Some(⊥() -> true
[error]    )' is not None (grammarFindingTest.scala:68)
[info] 
[info] minisat should
[error]  x say bottom is invalid as well
[error]   'Some(⊥() -> true
[error]    )' is not None (grammarFindingTest.scala:74)
```

frehn:
Thanks for pointing this out, Gabriel. The long overdue fix for this problem can be inspected in commit 9bbddb446518bcf15feda899244f177fc5d87261.

gebner:
Thanks Daniel; it still failed for me because the Tseitin CNF used HOL top and bottom, so I changed that in 65c901ac.

frehn:
Thanks, I missed that!

Ceterum censeo issue #292 delendam esse.

PR #331, opened 2015-05-27
Regular layers

gebner:
This PR implements a refactoring of the expression classes, adding traits to expressions as determined by a (more or less) regular tree automaton (hence the name).

In short:  We now consider a data type of lambda expressions, where some lambda expressions are automatically assigned more specific types, for example being a FOLTerm.  These more specific types are implemented as traits, and they are assigned centrally (in the [determineTraits](https://github.com/gapt/gapt/blob/regular-layers/src/main/scala/at/logic/gapt/expr/traits.scala#L43-130) object).

The upshot of this is that equal expressions will always have the same types; Bottom() will always have a type that is a subtype of FOLFormula, whether you create it in the context of a propositional formula, or in the context of a HOL formula.

Other minor changes:
 * Bottom() and Top() are now logical constants, and *not* atoms.  This should prevent bugs like #315 and #309.


mrukhaia:
Is this brunch already merged? Although the schemaprover brunch is not
finished yet, code compiles and tests are not failing, so I would like to
merge it in master before this brunch is merged.

Best regards
Mikheil Rukhaia

shetzl:
Dear Mikheil,

The schemaprover branch is not yet ready to be merged into master:

1) Please follow Daniel's suggestion to avoid code duplication by deriving SchemaProofStrategy from (at least) PropositionalProofStrategy but better even from a to be newly created FirstOrderProofStrategy.

2) In my commit b74ef1259684b26b1c86a8d5331986fd619831d9 from May 1 I improved the solver by changing the invariant from proving the goal sequent to proving a subsequent of the goal sequent. This improves performance for propositional logic by about a factor of 20. Merging this with the current code in schemaprover will require some work.

The best strategy for avoiding a too large gap to the master branch is to periodically (e.g. once a week) merge master into your feature branch.

3) There is a number of smaller things that first have to be cleaned up, e.g. the println in commit 580def550af220aae1b13333d7005d3140d03ab2 or the log message  debug("running solvePropositional") at solve.scala:46 which should read something like "running schema prover".

Do you have time to participate in the gapt meeting next Wednesday, 14:45 via skype?

Best regards,
Stefan

3)

mrukhaia:
Dear Stefan,

I'm not sure that I understood this sentence properly:

"The best strategy for avoiding a too large gap to the master branch is to
periodically (e.g. once a week) merge master into your feature branch."

Does it mean that I can merge master into schemaprover, not shemaprover to
master? I thought if I merge master into schemaprover, joining it back
would cause problems, when schemaprover brunch is finished.

I'll try my best to participate on the gapt meeting.

Best regards
Mikheil Rukhaia

shetzl:
Dear Mikheil,

Yes, I really mean merging master into schemaprover. This will not at all cause problems when you finally merge schemaprover into master, on the contrary: it will make it easier because each single merge will create less conflicts.

Have a look, e.g., on the history of the regular-layers branch on https://github.com/gapt/gapt/network. There we have also used this strategy.

Best regards,
Stefan

mrukhaia:
Nice, I did not know about it. Then it is not a problem, I'll merge master
into schemaprover and continue to work there.

Best regards
Mikheil Rukhaia

PR #332, opened 2015-05-30
Spring cleaning: Remove junit, specs2-mock, and scalacheck dependencies.

gebner:
As far as I can tell, these are no longer needed.
 * specs2-mock and scalacheck were only used in imports
 * junit was only used in the maven build.

Any objections to removing them?


quicquid:
Great idea - we don't have much data we could mock anyway.

Issue #318, opened 2015-04-08
Prover9 interface is sensitive to equation order


gebner:
```scala
import at.logic.gapt.provers.prover9.Prover9Prover
import at.logic.gapt.formats.prover9.Prover9TermParserLadrStyle.parseFormula

new Prover9Prover().getLKProof(FSequent(
    Seq("(all x (all y (all z (x*(y*z) = (x*y)*z))))",
        "(all x (e*x = x))",
        "(all x (all y (i(x)*x = e)))")
      map parseFormula,
    Seq(parseFormula("a*i(a) = e"))))
```

This fails with:
```
at.logic.gapt.proofs.resolution.algorithms.ResolutionException: Resolution Exception: Clause [ :- =(*(*(v0, v1), v2), *(v0, *(v1, v2)))] is not reflexivity and not contained in CNF(-s) [
 :- =(*(x, *(y, z)), *(*(x, y), z));
 :- =(*(e, x), x);
 :- =(*(i(x), x), e);
=(*(a, i(a)), e) :- 
]
	at at.logic.gapt.proofs.resolution.algorithms.PCNF$.apply(PCNF.scala:69)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$PCNFMemoTable.getPCNF(robinson.scala:24)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$$anonfun$apply$1.apply(robinson.scala:36)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$$anonfun$apply$1.apply(robinson.scala:36)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:82)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:170)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:130)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:170)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:130)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:129)
	at at.logic.gapt.proofs.resolution.algorithms.RobinsonToLK$.recConvert(robinson.scala:170)
[...]

PR #348, opened 2015-06-18
New Prover9 interface

gebner:
We've been having quite a few problems with the prover9 import failing because we couldn't adapt the prover9 proofs to our CNF, see #318 for example.

The mismatch between the initial clauses of the imported prover9 proof and our CNF clauses should be fixed by `fixDerivation`, however particularly in the presence of equations `fixDerivation` quite often fails to fix this mismatch.

The previous plan was to improve fixDerivation to find a larger class of derivations, e.g. find the initial clause used by prover9 using propositional resolution + variable renaming.  But after yesterday's discussion with Alex, this seems to be a hard problem with no obvious solution in the presence of symmetric predicates (equality).

This PR changes the way we call prover9 instead:

 * Prover9 gets a list of clauses, not a quantified formula.
 * Since prover9 gets explicit clauses, it does not need to do a CNF and references our clauses in the proof output.  This makes fixDerivation unnecessary.

There are some open issues:
 - [x] What is the grounding in the current prover9 interface used for?  Should it be implemented here as well?
 - [ ] We don't even do Skolemization?  Should we do it by default?
 - [x] The prover9 import for existing prover9 output is not implemented yet.  Unfortunately the change to direct CNF output does not improve the situation with `fixDerivation` here as it is still needed.
 - [x] Is there a nicer way to apply substitutions to resolution proofs instead of calling `RenameResproof`?

@shetzl @frehn Any comments?


quicquid:
* RenameResproof only does renaming/replacement of subterms (in particular it reverts the renaming introduced by tptp_to_ladr), not substitution. It does not prevent you from replacing variables, but in general, the approach is only sound when replacing ground terms by other ground terms(this is also written in the comment).

* Afaik substitution for resolution is only explicitly represented by the Instance rule (Robinson2LK only substitutes LKProofs) and implicitly by the unifier passed as an argument to the other rules.

gebner:
Thanks!  I originally thought the variables in the initial clauses of the resolution proof must have the same names as in the end-sequent in order for RobinsonToLK to work--but RobinsonToLK seems to handle variable renamings just fine.

gebner:
> What is the grounding in the current prover9 interface used for?

Apparently in SipProver.toLKProof (in the inductive prover) we want to prove sequents with free variables.

quicquid:
Yes, what Robinson2LK (and CERES does) is to find an input clause which subsumes the clause (via a substitution sub) we want to project to. Then it applies the sub to the CNF projection and weakens the missing literals. The reason is that the prover might reorder the literals (since they're just a set).

src/main/scala/at/logic/gapt/cli/GAPScalaInteractiveShellLibrary.scala
```diff
@@ -300,21 +305,18 @@ object refuteFOLI {
 
 object prover9 {
   //we have to refute
-  def apply( filename: String ): Option[RobinsonResolutionProof] = Prover9.refute( filename )
+  //  def apply( filename: String ): Option[RobinsonResolutionProof] = new Prover9Prover().parseProof( filename )
```
gebner: This method (and refuteTPTP) is not used anywhere in the gapt source tree.  Does anybody want it back?

src/test/scala/at/logic/gapt/proofs/ceres/acnfTest.scala
```diff
@@ -95,7 +96,7 @@ class acnfTest extends Specification {
 
       //for (p <- proj) println(p)
       val rlkp = RobinsonToLK( rp.get )
-      val gproj = proj map ( applySubstitution( _, FOLSubstitution( ( u, b ) :: Nil ) )._1 )
+      val gproj = proj map ( applySubstitution( _, FOLSubstitution( u -> a, v -> a ) )._1 )
```
gebner: @quicquid I had to change to grounding substitution here in order to make the projections fit.  Is this a problem?  Does the grounding depend on which resolution proof we get?

src/main/scala/at/logic/gapt/cli/GAPScalaInteractiveShellLibrary.scala
quicquid: I just checked, it isn't necessary anymore. The only command which used it was loadProver9Proof in the CLI, but this already calls parse_prover9.

src/test/scala/at/logic/gapt/proofs/ceres/acnfTest.scala
quicquid: The ACNF code is a leftover from the schema code. Your commit mentioned that the proofs prover9 returns can change, then the ground substitution can also change. I'm actually not sure what to do with the whole file - we have a working LK version in CERES and a LKskc version in ceres_omega. The only reason to keep ACNF.scala is the schema specific code.

gebner:
Running the regression tests on the compile server revealed the need for the strong quantifier special casing in the TSTP import.  This is now implemented.

Local testing now shows no regressions compared to master (some timeouts have become OutOfMemoryErrors though).

PR #373, opened 2015-08-24
Remove weakening in PCNF?

gebner:
When we convert resolution proofs to LK, the proofs often begin with an axiom, weaken it with an extra copy of the formula, and then contract the copy away.  This seems to be caused by a call to `WeakeningMacroRule` in the PCNF code.

I'm not entirely sure what the weakening is supposed to do, does anybody have any ideas?

Removing the weakenings does not cause any tests to fail.


gebner:
Let's see what the regression tests say.

gebner:
The regression tests didn't uncover any problems with this change.

PR #379, opened 2015-09-11
Remove GAPScalaInteractiveShellLibrary.

gebner:
The `GAPScalaInteractiveShellLibrary` has historically been intended as the "user-visible" layer of gapt that is to be used in the CLI.  However this approach has had several issues:

* There are lots of functions in `GAPScalaInteractiveShellLibrary` that just rename existing functions.  These names are different for no apparent reason (`elminateCuts` vs. `ReductiveCutElim`).
* Sometimes they are even duplicated (`parse.fol` vs `parseFOL`)
* In general, they are not as well maintained as the code they wrap.
* Using these wrappers in the user's guide creates an unnecessary burden for newcomers who want to start developing.
* Helping newcomers is really difficult if they want to follow the user's guide and you don't recognize any functions they want to use...

In this pull request, I have removed the `GAPScalaInteractiveShellLibrary`.
* Functionality that was only readily available via the wrappers has been moved into the corresponding packages (`prooftool`, `parseFOL`, ...)
* For wrappers that just renamed other functions, I have updated the user's guide to use the name of the existing functions.
* I have ported all the scripts to the new names (except `CutIntroToolBox.scala`, which I hasn't worked for a long time now.)
* The user's guide is updated. 

Some issues remain:
* `equation_example`, `philsci`, `tbillc` should get better names
* What is [`css`](https://github.com/gapt/gapt/commit/ed0a43013a7a4677a8665c10bb0ce2e417d28458#diff-b52d7a31ec47281ba63c6a67de9eb296R12)?


quicquid:
css stands for characteristic sequent set

PR #449, opened 2015-12-09
Unary eq

quicquid:
Moves ceres, ceres omega and llk to the new lk. Ceres omega is not working properly yet and ceres has some proofs not fully working, but the tape and lattice proof test go through.


src/main/scala/at/logic/gapt/proofs/lkNew/lkNew.scala
```diff
@@ -271,7 +271,7 @@ object Axiom {
    * @return An axiom of the appropriate type, depending on sequent.
    */
   def apply( sequent: HOLSequent ): InitialSequent = sequent match {
-    case Sequent( Seq( f: HOLAtom ), Seq( g: HOLAtom ) ) if f == g => LogicalAxiom( f )
+    case Sequent( Seq( f: HOLFormula ), Seq( g: HOLFormula ) ) if f == g => LogicalAxiom( f )
```
gebner: You can drop the type matcher here.

src/main/scala/at/logic/gapt/proofs/lkNew/lkNew.scala
```diff
@@ -1014,7 +1014,7 @@ object ForallLeftRule extends ConvenienceConstructor( "ForallLeftRule" ) {
 
     mainFormula match {
       case All( v, subFormula ) =>
-        val auxFormula = Substitution( v, term )( subFormula )
+        val auxFormula = BetaReduction.betaNormalize( Substitution( v, term )( subFormula ) )
```
gebner: Thanks for looking over the new LK w.r.t. higher-order proofs.  I was afraid some problems would surface sooner or later.

src/main/scala/at/logic/gapt/proofs/lkNew/LKToExpansionProof.scala
```diff
@@ -21,7 +21,9 @@ object LKToExpansionProof extends Logger {
   def apply( proof: LKProof ): ExpansionSequent = proof match {
 
     // Axioms
-    case LogicalAxiom( atom )                    => Sequent( Seq( ETAtom( atom ) ), Seq( ETAtom( atom ) ) )
+    case LogicalAxiom( atom: HOLAtom )           => Sequent( Seq( ETAtom( atom ) ), Seq( ETAtom( atom ) ) )
+
+    case LogicalAxiom( formula )                 => apply( AtomicExpansion( formula ) )
```
gebner: I'm not sure this is correct.  AFAIK the expansion sequent extraction requires regular proofs (at least to be correct), so we should probably have a `regularize(AtomicExpansion(proof))` somewhere.

src/main/scala/at/logic/gapt/proofs/resolution/RobinsonToRal.scala
```diff
@@ -22,11 +23,17 @@ abstract class RobinsonToRal {
...
-    case _: InitialClause             => RalInitial( p.conclusion map convert_formula map { Set[LambdaExpression]() -> _ } )
+    case _: InitialClause             => RalInitial( p.conclusion map convert_formula map { Seq[LambdaExpression]() -> _ } )
     case Factor( p1, i1, i2 )         => RalFactor( apply( p1 ), i1, i2 )
     case Instance( p1, subst )        => RalSub( apply( p1 ), convert_substitution( subst ) )
     case Resolution( p1, i1, p2, i2 ) => RalCut( apply( p1 ), Seq( i1 ), apply( p2 ), Seq( i2 ) )
     case Paramodulation( p1, eq, p2, lit, pos, dir ) =>
-      RalPara( apply( p1 ), eq, apply( p2 ), lit, pos, dir )
+      eq match {
+        case Ant( _ ) =>
+          throw new Exception( "Sequent index of equality in para rule must be in succedent!" )
```
gebner: Since this is a case that can never happen (by construction of `Paramodulation`), I'd rather not handle it and just add the `@ Ant(_)` to the Paramodulation pattern match--that keeps the code shorter more to the point.  But this is probably personal preference.

src/main/scala/at/logic/gapt/provers/prover9/prover9.scala
```diff
@@ -67,9 +67,10 @@ class Prover9( val extraCommands: ( Map[Const, Const] => Seq[String] ) = _ => Se
       } )( formula )
   private def toP9Input( clause: HOLClause ): String = toP9Input( renameVars( clause.toFormula ) )
   private def toP9Input( expr: LambdaExpression ): String = expr match {
+    case Top()                => "$T"
```
quicquid: What about Prover9.refute(List(HOLSequent(Top()::Nil, Nil) ) ? (Haven't tried it, but that was the intention).

src/main/scala/at/logic/gapt/proofs/resolution/RobinsonToRal.scala
quicquid: Yes, I also saw that you were experimenting with adding the narrower type in the constructor of Ral. I guess we could make the narrower type more uniform over all calculi.

src/main/scala/at/logic/gapt/proofs/resolution/RobinsonToLK.scala
```diff
@@ -85,17 +85,20 @@ object RobinsonToLK {
...
-          ParamodulationRightRule( f( p1 ), p1.conclusion( eq ), f( p2 ), p2.conclusion( lit ), p.rewrittenAtom ),
+          ParamodulationRightRule( f( p1 ), p1.conclusion( eq ),
+            f( p2 ), p2.conclusion( lit ), p.rewrittenAtom ),
           endSequent ++ p.conclusion, strict = false
         )
     } )
-
-    WeakeningContractionMacroRule( f( resolutionDerivation ), endSequent )
+    val rproof = f( resolutionDerivation )
+    println( s"endSequent: $endSequent\n$rproof\n" )
```
quicquid: Argl overlooked one :(

src/main/scala/at/logic/gapt/proofs/lkNew/LKToExpansionProof.scala
quicquid: Hm, I'll look it up - if it needs to be regular, it is better to call the atomic expansion before the et extraction because regularization is not local to the subproof.

src/main/scala/at/logic/gapt/provers/prover9/prover9.scala
gebner: Ah, I can see where the misunderstanding comes from: Top is not an atom, it is a logical connective.  Hence, it cannot occur in a HOLClause.

quicquid:
Should we merge despite the 0.8% decrease in coverage?

gebner:
If I see this correctly, we can remove the old LKsk implementation after we merge this branch.  This should more than make up the -0.8%. :smile:

Please merge!

quicquid:
I'll check if this is really the case (the schema code still contains a lot of the old implementation), but i'm pretty confident that we can get rid of it soon if not immediately.

PR #459, opened 2015-12-17
New Vampire interface

gebner:
* Works with the CASC version of Vampire 4.0.
* Supports proof import
* Uses the same installation check as eprover, prover9, etc. and should hence also work on Windows


gebner:
Merging because this solves the Vampire bug on Windows.

PR #500, opened 2016-03-23
Skolem symbol validation

gebner:
This PR adds validation for Skolem symbols using so-called Skolem definitions: a Skolem definition is similar but slightly different from defining Skolem symbols using the epsilon operator:

Syntactically it is a map `s_i → λx_1 .. λx_n Qy φ(x_1, .., x_n, y)`, where Q is a quantifier.
Then `s_i(x_1, .., x_n)` is the Skolem term used for the formula `Qy φ(x_1, .., x_n, y)`, where Qy is strong.

This Skolem term corresponds to the epsilon term `εy φ(x_1, .., x_n, y)` or `εy ¬φ(x_1, .., x_n)`, depending  on whether Q is ∃ or ∀.  The reason we don't use epsilon terms directly is that this makes it impossible to deskolemize a formula based on just the Skolem definitions alone: for example both `∃x ∀y φ` and `∃x ¬∃y¬ φ` would define their Skolem functions using the same epsilon terms.

@quicquid In order to make this work, I had to add an extra argument to some of the Skolem functions in the nTape proofs (code-wise this is added in the LKToLKsk function); this is how the definitions look like for nTape3:
```
scala> println(nTape3.expansion_proof.skolemFunctions)
s_10:i>(i>o)>(i>i)>i → λ'\\sigma' λY λh ∀i (i < s_5(Y:i>o) + 1 + 1 ⊃ f(h(i)) = '\\sigma')
s_4:i>(i>o)>(i>i)>i → λ'\\sigma' λ(Y:i>o) λh ∀i (i < 0 + 1 ⊃ f(h(i)) = '\\sigma')
s_2:i>i>i → λx λy ∃k x + k + 1 = y
s_1:(i>o)>(i>i)>i → λY λh ∀j (s_0(Y:i>o, h:i>i) < 0 + 1 ∧ (j < 0 + 1 ∧ s_0(Y, h) < j) ⊃
      h(s_0(Y, h)) < h(j))
s_0:(i>o)>(i>i)>i → λ(Y:i>o) λh ∀i ∀j (i < 0 + 1 ∧ (j < 0 + 1 ∧ i < j) ⊃ h(i) < h(j))
s_8:(i>o)>(i>i)>i → λY λh ∀j (s_7(Y:i>o, h:i>i) < s_5(Y) + 1 + 1 ∧
        (j < s_5(Y) + 1 + 1 ∧ s_7(Y, h) < j) ⊃
      h(s_7(Y, h)) < h(j))
s_6:i>(i>o)>i>i → λ'\\sigma' λY ∃h (∀i ∀j (i < s_5(Y:i>o) + 1 ∧ (j < s_5(Y) + 1 ∧ i < j) ⊃
          h(i) < h(j)) ∧
      ∀i (i < s_5(Y) + 1 ⊃ f(h(i)) = '\\sigma'))
s_15:(i>i)>i>i → λh λs ∀i (i < s_11 + 1 ⊃ f(h(i)) = s)
s_7:(i>o)>(i>i)>i → λY λh ∀i ∀j (i < s_5(Y:i>o) + 1 + 1 ∧ (j < s_5(Y) + 1 + 1 ∧ i < j) ⊃
      h(i) < h(j))
s_5:(i>o)>i → λY ∀n (Y(n) ⊃ Y(n + 1))
s_14:(i>i)>i → λh ∀j (s_13(h:i>i) < s_11 + 1 ∧ (j < s_11 + 1 ∧ s_13(h) < j) ⊃
      h(s_13(h)) < h(j))
s_13:(i>i)>i → λh ∀i ∀j (i < s_11 + 1 ∧ (j < s_11 + 1 ∧ i < j) ⊃ h(i) < h(j))
s_11 → ∀n ∃h (∀i ∀j (i < n + 1 ∧ (j < n + 1 ∧ i < j) ⊃ h(i) < h(j)) ∧
    ∃s ∀i (i < n + 1 ⊃ f(h(i)) = s))
s_12:i>(i>o)>i>i>i → λ'\\sigma' λ(Y:i>o) λn ∃h (∀i ∀j (i < n + 1 ∧ (j < n + 1 ∧ i < j) ⊃
          h(i) < h(j)) ∧
      ∀i (i < n + 1 ⊃ f(h(i)) = '\\sigma'))
s_3:i>i → λx ∃k x = k + 1
```
For s_4, s_6, s_10, s_12 the Skolem functions get an extra `\sigma` as this is a free variable in the formula with the strong quantifier in the proof.  This free variable is already present in the second-order argument: for example s_4 was used as `s_4(some_formula_containing_sigma)`, now it is `s_4('\\sigma', some_formula_containing_sigma)`.  Unfortunately there is no lambda term `M` such that `M(some_formula_containing_sigma) = '\\sigma'`, so that's where the extra argument comes from.

### DIFF
--- a/core/src/main/scala/at/logic/gapt/expr/constructors.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/constructors.scala
@@ -89,6 +89,8 @@ class QuantifierHelper( val q: QuantifierC ) {
 object All extends QuantifierHelper( ForallC )
 object Ex extends QuantifierHelper( ExistsC )
 
+object Quant { def unapply( f: HOLFormula ) = All.unapply( f ) orElse Ex.unapply( f ) }
+
 class BinaryPropConnectiveHelper( val c: MonomorphicLogicalC ) {
   def apply( a: LambdaExpression, b: LambdaExpression ): HOLFormula =
     Apps( c(), a, b ).asInstanceOf[HOLFormula]
--- a/core/src/main/scala/at/logic/gapt/expr/hol/SkolemFunctions.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/hol/SkolemFunctions.scala
@@ -1,0 +1,44 @@
+package at.logic.gapt.expr.hol
+
+import at.logic.gapt.expr._
+import at.logic.gapt.proofs.expansion.linearizeStrictPartialOrder
+
+/**
+ * List of definitions of Skolem symbols.
+ *
+ * A Skolem definition is similar but slightly different from the epsilon operator:
+ *
+ * Syntactically it is a map s_i → λx_1 .. λx_n Qy φ(x_1, .., x_n, y), where Q is a quantifier.
+ * Then s_i(x_1, .., x_n) is the Skolem term used for the formula Qy φ(x_1, .., x_n, y), where Qy is strong.
+ *
+ * This Skolem term corresponds to the epsilon term εy φ(x_1, .., x_n, y) or εy ¬φ(x_1, .., x_n),
+ * depending on whether Q is ∃ or ∀.  The reason we don't use epsilon terms directly is that this makes
+ * it impossible to deskolemize a formula based on just the Skolem definitions:
+ * for example both ∃x ∀y φ and ∃x ¬∃y¬ φ would define their Skolem functions using the same epsilon terms.
+ */
+case class SkolemFunctions( skolemDefs: Map[Const, LambdaExpression] ) {
+  skolemDefs foreach {
+    case ( s, d @ Abs.Block( vs, Quant( v, f ) ) ) =>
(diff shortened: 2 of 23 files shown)


gebner:
Forgot to mention this, but all tests pass.  I'm just afraid that it might cause some problems in the nTape code that is not tested, or some other problems down the line in the future.

quicquid:
Regarding the change for nTape / LKsk - the quantifier rules now differ from the ones defined in the paper, but I think it should be treatable. If a proof holds for all interpretations of s(x, t[x]), then it should also hold for all interpretations of s'(t[x]). We don't have (relative) completeness of CERESomega yet, where the roles might be reversed.

gebner:
OK, thanks for checking this.

PR #504, opened 2016-04-07
Spring cleaning

gebner:
Like every year around this time, the temperatures are climbing, and I suddenly get in the mood to clean up old and abandoned parts of gapt.  This time I propose to delete 25k lines of code.

* The biggest and probably most controversial part is the schema code.  I've been thinking long and hard about this, but I have come to the conclusion that keeping it in the tree serves no useful purpose.
  * The schema code has no maintainer.  Even worse, nobody knows what it should do, and whether it actually still works.
  * When Martin was in Vienna last time, he told me that the `IntVar` matcher had been silently broken for a year (since the merge of the regular layers branch).
  * Code relating to schemas is all over the place: in prooftool, latex exporting, hol2fol, etc.  This makes it difficult to refactor these parts of the code, as there's the possibility of breaking the schema parts.
  * If anybody wants to work on schema in the future, they are probably much better off writing the schema support from scratch, instead of trying to understand and fix bugs in an old and unmaintained code base.  I *have* tried to refactor at least the expression-level code in the regular layers branch, and given up because of the confusing code.
* The old LKsk.  Martin, I understand that you do not want to remove it lightheartedly if the new LKsk doesn't work with your thesis yet (does it now?).  However, the expansion tree extraction in the old LKsk is pretty much broken now, so I don't think it is very useful.
* The old LLK.  There is nice shiny new one in the `llkNew` package.
* The old delta-table algorithm.  Again, we have a rewritten version.
* Some formatting functions.

If there's anything else you'd like to never see again, please tell me!  Also if you absolutely want to keep something!


quicquid:
I guess we can also retire the diophantine equation solver (AC unification was already removed i think). The combination of disjoint theories was never implemented, so a pure AC unification algorithm is pretty useless. If I understand it correctly, superposition based provers approach this differently anyway.

gebner:
* Stubs for SymbolA are gone as well, possible due to the removal of the schema code.

gebner:
> If I understand it correctly, superposition based provers approach this differently anyway.

I researched this a bit when I added AC support to Escargot.  From what I can tell, E, Prover9, and SPASS do nothing more than:

 * Unit rewriting with the associativity, i.e. associate everything to the left (or right, depending on the term order).
 * Conditional unit rewriting with the commutativity clause, i.e. simplify using commutativity whenever it results in a smaller term.
 * Conditional unit rewriting with x+(y+z)=y+(x+z).  This implements a version of bubble sort (at least if the terms are comparable).
 * A simple AC redundancy check that rejects obviously redundant newly derived clauses such as x+(y+z)=(z+x)+y.  (This is pretty crucial because otherwise you're inundated with useless junk.)

AFAICT none of the general purpose first-order provers implements general AC unification or matching.

gebner:
The change in code coverage is impressive: +10%!

gebner:
TODO: remove `FOLUnificationAlgorithm`

PR #518, opened 2016-04-18
Spring cleaning, part 2.

gebner:
I have just noticed that we have a needlessly complex implementation of finitely branching trees.  It seems that it is only used in prooftool to show CERES structs.  Of course, I'd like to get rid of it.

@quicquid Are you using prooftool to view structs?

If anybody wants to keep the prooftool tree viewer, then we should:

 1. Make it possible to use `prooftool(someStruct)`.  Currently, you have to some dance with `structToExpressionTree`.
 2. Get rid of this tree data structure, and use lambda expressions instead.  Then we could also use the viewer for other kinds of expressions.


quicquid:
I wouldn't miss it too much, but from time to time, the struct viewer can be quite useful, especially in cases when the resulting clause set becomes too big to compute. I think the original idea of the tree classes was that we could show arbitrary proof trees (e.g. raw tstp), but structs only have additional information on the leaves and need a bit different handling. Replacing that code by an arbitrary term tree viewer sounds good, if there is interest from other people too. If not, the easiest way to keep the functionality is probably to replace DrawTree by a specialized DrawStruct which works directly on the structs.

PR #521, opened 2016-05-05
LK: add Skolem inferences.

gebner:
This PR adds two inferences to LK: ∀sk:r and ∃sk:l (only showing one)
```
Γ :- Δ, φ(s(...))
-------------------  ∀sk:r
Γ :- Δ, ∀x φ(x)
```

 * This makes LK more symmetric to our expansion tree calculus.
 * `SomeProver.getLKProof` no longer fails on strong quantifiers.
 * `ExpansionProofToLK` can now handle all expansion trees.
 * It is now easy to compare `LKToExpansionProof(RobinsonToLK(p))` and `RobinsonToExpansionProof(p)`.
  * The difference is particularly striking when comparing the expansion proofs in the case of structural clausification.
 * It is actually remarkably little effort to add new inferences.

### DIFF
--- a/core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
@@ -124,6 +124,10 @@ class ExpansionProofToLK(
         mapIf( solve( cuts, expSeq.updated( i, f ) ), f.shallow, false ) { ExistsLeftRule( _, sh, ev ) }
       case ( ETStrongQuantifier( sh, ev, f ), i: Suc ) =>
         mapIf( solve( cuts, expSeq.updated( i, f ) ), f.shallow, true ) { ForallRightRule( _, sh, ev ) }
+      case ( ETSkolemQuantifier( sh, skT, skD, f ), i: Ant ) =>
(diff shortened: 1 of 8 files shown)


quicquid:
Just a question: it seems LK and LKsk are converging - if the experiences go well here, it might be interesting to merge them completely. What do you think about it?

gebner:
That was my unspoken hope.  Thanks for actually bringing it up, and not seeing any obvious problems.

Come to think of it, adding Skolem inferences to LK in order to support LKsk was actually one of the original plans (half a year when we did the new LK).  The labels would then have been computed on the fly.  However, Daniel's PhD thesis and the paper were so convincing in their presentation of labelled sequents that the idea got dropped.

Having only a single LK-like calculus is probably a good idea for the long-term maintainability of CERESw.  But what I'd actually like to merge is resolution and Ral: I have for a long time played with the idea of having an "end-to-end" refutational calculus in GAPT, i.e. a "resolution" calculus that starts from the input formula/sequent--in some sense that's very similar to the TPTP proof output adopted by E/Vampire/metis.

With a bit of luck, this calculus could nicely represent structural clausification and allow easy expansion proof extraction while getting around the exponential blow-up (on the propositional level) when going to LK.  Having built-in Skolemization and clausification might also allow us to reconstruct and represent the preprocessing steps performed by first-order provers.  Depending on how easy it would be to extract expansion proofs, we could even have a more principled representation of splitting inferences.

PR #523, opened 2016-06-06
[RFC] New resolution calculus

gebner:
I do not intend to merge this very soon, at the very least I need to add some documentation.  I highly welcome any comments, criticisms, wishes, etc.!

Our current resolution calculus has several shortcomings, from my point of view the main ones are the following:
 1. Preprocessing steps, introduced definitions, etc. are not included or represented in the proof.  In order to convert resolution proofs to LK or expansion proofs, we have to pass the original end-sequent, the introduced definitions, and justifications for the input clauses (i.e. projections of the end-sequent) around.  This makes working with resolution proofs cumbersome.
 2. Splitting inferences are handled in an intransparent and confusing way.  In particular, we do not support Vampire's Avatar inferences.
 3. Due to 1., the current resolution calculus is only first-order.  As alluded to in [another pull request](https://github.com/gapt/gapt/pull/521#issuecomment-217735987), it might be a good idea to use the standard calculi for CERESω--this PR might pave the way towards unifying LK/LKsk and Ral/resolution.

Recently, I had the pleasure of being locked 10 hours inside an airplane during daytime and could thus try out some ideas that fix all three problems in order to pass the time.  The result is the new resolution calculus implementation in this PR.  Features:
 * The clausification is completely represented in the calculus.  Similar to higher-order resolution calculi (such as Ral), there are inferences that decompose formulas from the end-sequent into clauses.  This eliminates the mysterious `(cnf, justifications, definitions)` triples that were being passed around.
 * Splitting is supported through special rules for Avatar.  (SPASS-style splitting can be simulated as well, this works with the built-in SPASS interface.)   Hence the calculus does not reason about clauses, but A-clauses.  An A-clause is a pair of a clause and a conjunction of ground literals--the "assertion".  We write `l_1 ∨ l_2 ∨ ... ∨ l_n  ←  a_1 ∧ ... ∧ a_n` for an A-clause consisting of a clause `l_1 ∨ l_2 ∨ ... ∨ l_n` with assertion `a_1 ∧ ... ∧ a_n`.  An A-clause is interpreted as `a_1 ∧ ... ∧ a_n ⊃ l_1 ∨ l_2 ∨ ... ∨ l_n`.  Inferences such as resolution and paramodulation only operate on the clause part, not the assertion.
 * The resolution calculus is higher-order.
 * Expansion proof extraction works via a bottom-up expansion tree propagation.
 * Escargot is now a bit less sluggish and does Avatar splitting.

Obligatory prooftool screenshot:
```scala
prooftool(Escargot getRobinsonProof hof"∀x p x ⊃ ∀x p (f x)")
```
Sample proof in the new resolution calculus

To do:
 * When `ResolutionToLKProof` encounters a rule for a propositional connective, it converts the resolution proof to an expansion proof, and the expansion proof to LK.  The missing part is to transform a proof of e.g. `Γ :- a ∧ b` into a proof of `Γ :- a`--this should be possible without introducing a cut, by modifying the inferences that introduces `a ∧ b`.  This is not implemented yet because 1) these are 17 similar transformations that I do not want to write using copy&paste, and 2) our current LK visitor code does not track ancestorship data.
 * Splitting elimination.  The conversion of splitting inferences to LK creates Π_1-cuts.  
 * `ResolutionToLKProof` uses definition rules.  In some cases these cannot be eliminated easily: for example if CERES plugs in projections at the top which already contain other definitions.  The elimination fails because we cannot selectively eliminate definitions at the moment.
 * Documentation.
 * Actual proof import from Vampire.
 * The old `structuralCNF` clausification procedure that provides projections as justifications is still present.  It has the same functionality as `structuralCNF3`, but the unconditional construction of resolution proofs might cause a performance loss; I still need to test this.


gebner:
Splitting elimination, Vampire proof import, and documentation is done now.

I cleaned up the old clausification code to use in the SAT/MaxSAT solvers, as it is about 6x faster on the grammar minimization formulas.

gebner:
Merging this for testing in gapt 2.2.

PR #544, opened 2016-07-14
Remove LKsk and Ral.

gebner:
* All of the functionality of LKsk and Ral are included in the normal LK and resolution calculus nowadays (except for the labels).
 * Higher-order CERES works in the same way as before, just with LK instead of LKsk, and resolution instead of Ral.  Some new features might become possible due to the switch: conversion of the built-in clausification rules to LK is implemented in the normal resolution calculus, but not in Ral AFAICT.
 * LKsk and Ral are not extensively tested since they are only used by CERESω.  Some parts of Ral--for example the built-in clausification--are even completely unused and untested; by contrast, this part of the resolution calculus is heavily tested as it is essential to proof import and the resolution prover interfaces.
 * There are also subtle mismatches between e.g. LK and LKsk: LK supports theory axioms, while LKsk does not.  As a consequence, `LKToLKsk` fails on some proofs.  This is no longer the case: Skolemization is total now (modulo bugs).
 * We have 3500 lines of code less to maintain.

@quicquid Tell me when you're done with your thesis, I don't want to cause any delays.


gebner:
Merged.

quicquid:
Just a question, can you leave this branch open for a while? The removed projections computation was more general  and is the only one which handles CERES omega proofs correctly. Right now it is impossible to create a labelled characteristic sequent set anymore - in fact I don't see any labels computed at all.

gebner:
Ok, this is too late now as I've already merged it into master.  These objections would have been helpful earlier, this PR has been open for almost half a year now.
The labels are still there in the 2.3 release however.

> The removed projections computation was more general and is the only one which handles CERES omega proofs correctly.
Can you point me to an example?  All the tests work fine.

> It is quite easy to provide a refutation with wrong skolem terms which will then create a faulty LKsk proof without errors.
Just to clarify, there should not be any *soundness* problems.  The issue here is that it becomes easier to produce proofs that cannot be deskolemized in linear time, and which probably cannot be easily or efficiently transformed to cut-free proofs (except of course via cut-elimination).
Hmm, right now I see no trivial solution--again, these comments would have been better placed earlier.  We can a) revert this PR, or b) add functions that compute the labels.

quicquid:
> Ok, this is too late now as I've already merged it into master. 
I was waiting to have a proper look for this after I submitted my thesis (which is still delayed, i hope ill manage to have the final draft this week). I had only realized the removal of LKsk - which is fine because the label information can be computed - but the removal of Ral.

> Can you point me to an example? All the tests work fine.

The main problem of CERESw is that there is no prover where we can influence the skolem terms created during dynamic CNF transformation. For my thesis I focused on the subset without CNF transformations i.e. function variables. Since all the tests are in that subset, they go through.

The point where the method breaks is when we translate the Ral quantifier rule:

    ∀F, ... ⊢ ...
    ---------------
    F(S), ... ⊢ ...

 back into LKskc. Essentially we apply the inverse transformation to the succedent of the sequent in the axiom ∀F ⊢ ∀F to obtain:

    F(S*) ⊢ F(S)
    --------------
    ∀F ⊢F(S)

(This is essentially performing the skolemization step at the leaf of the proof). S* now must agree with the skolem context S from the LKsk proof, but the only way to control the shape of S is via the labels.

Because of the non-existing automation, supporting this rule was always a low priority for me, that's why it wasn't implemented before. I did prepare the implementation by adding tracing code to find the axiom which needs to extended and to pick an auxiliary formula with the correct label (thats what pickrule does). With  the move to the new resolution calculus, the unsupported CNF transformations are now possible to perform, but the labels are unchecked i.e. the method doesn't throw an exception when the labels don't fit.

> Hmm, right now I see no trivial solution--again, these comments would have been better placed earlier. We can a) revert this PR, or b) add functions that compute the labels.

As mentioned above, I had postponed this because I knew I had to thoroughly review the code not to miss any problems. It's clear you can't wait forever so I postponed for too long :( What I can do quickly is to write a test which lets CERESomega fail on inputs which might suffer from this problem: this is basically every proof with strong quantifiers or a predicate variable in an axiom with cut-ancestors(*). That would avoid a revert and let us add the label computation whenever there is energy / time.

How much more effort to put into this depends on the plans for the future. I thought about experimenting with the full CERES omega method -- to finally get the higher-order Fürstenberg proof back into the system -- as soon as I had a little breath, but at least for the next year, this will only be a side-project.

The main problem blocking publications about such experiments is the lack of a higher-order resolution prover with control over the skolem symbols. An alternative might be a manual specification via a subset of gaptic - i'd say Ral is closer to tableaux with atomic cuts than to resolution. In both cases, this is more than I can do on my own. If the gapt team decides that the fragment beyond the descriptions of (*) is interesting for a future student or other member of the team, I'd implement proper label computation and salvage the tracing and label checking code. For my thesis, it is probably the safest to point to the tag of gapt 2.3.

gebner:
Thank you for the long and thoughtful response!

> I was waiting to have a proper look for this after I submitted my thesis (which is still delayed, i hope ill manage to have the final draft this week).

I wish you the best for your thesis!

> What I can do quickly is to write a test which lets CERESomega fail on inputs which might suffer from this problem: this is basically every proof with strong quantifiers or a predicate variable in an axiom with cut-ancestors(*).

Strong quantifiers in the input proof should not be a problem, we don't use any sophisticated Skolemization there.  Really the only potential problem are Skolem inferences in the resolution refutation, and they don't occur in your thesis or in any of the example proofs, right?

Please don't feel pressured to hurry anything.  For now I think we can leave it as it is, in particular given that we cannot generate these refutations automatically anyhow, and all the hand-written ones are right.

quicquid:
Thanks for the encouraging words :-) Things related to my thesis will work out, I crafted the problem such that it is untouched by these complications -- in some variants, I have a free predicate variable, but no HO-ATP can solve the problem anyway, even when labels are ignored. I only need to document a specific version of gapt and 2.3 is a good point for that.

PR #600, opened 2017-01-26
Added ACNF and ACNF_Top extraction to Reductive cut-elimination.

Ermine516:
To reductive cut elimination will still run is you type the following
command: ReductiveCutElimination(Proof). However, if you add a context,
ReductiveCutElimination(Proof,Context) it will also elimination all definition.
This will be reworked in an up coming commit to eliminate definitions
as part of the cut elimination proceedure. If you add a context or leave
it null and add a true you will eliminate to the first ACNF reached, i.e.
ReductiveCutElimination(Proof,Context,true) or
ReductiveCutElimination(Proof,null,true). Adding a second true argument will
eliminate to ACNF_top, i.e. ReductiveCutElimination(Proof,Context,true,true) or
ReductiveCutElimination(Proof,null,true,true). The following commands will
disable cleaning of structural rules:
ReductiveCutElimination(Proof,null,true,true,false)  //ACNF_top no def elim
ReductiveCutElimination(Proof,null,true,false,false) //ACNF no def elim
ReductiveCutElimination(Proof,null,false,false,false) //cutfree no def elim
ReductiveCutElimination(Proof,context,*,*,false) // def elim


core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
-   * rules. If the call returns false, the traversion continues.
-   *
-   * After every application of a reduction, pred_done(global) is called.
-   * If it returns true, the algorithm terminates, returning the current proof.
-   * If it returns false, the algorithm continues to traverse the proof.
-   *
-   * This means that pred_cut and pred_done allow the definition of a (not necessarily
-   * terminating!) cut-elimination strategy. A standard implementation (reducing
-   * left-uppermost cuts until the proof is cut-free) is provided by another
-   * apply method in this class.
```
shetzl: Please don't delete scaladoc comments. They are very helpful for future users of the code.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -29,7 +30,10 @@ object ReductiveCutElimination {
    * @param proof The proof to subject to cut-elimination.
    * @return The cut-free proof.
    */
-  def apply( proof: LKProof ) = new ReductiveCutElimination().eliminateAllByUppermost( proof )
+  def apply( proof: LKProof, ctx: Context = null, ACNF: Boolean = false, Top: Boolean = false, CleanStructRules: Boolean = true ) =
```
shetzl: If you add arguments to a high-level function like this, please document what they do in the scaladoc comment above.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
+  def InitAx( proof: LKProof ): Boolean = proof match {
+    case LogicalAxiom( _ ) => true
+    case _                 => false
+  }
 }
 
 class ReductiveCutElimination {
   val steps = mutable.Buffer[LKProof]()
-  var recordSteps: Boolean = false
+  var recordSteps: Boolean = true
```
shetzl: What is the reason for changing the default value here?

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
     pr
   }
 
+  def elimToACNFTop( proof: LKProof, ctx: Context, CleanStructRules: Boolean = true ): LKProof =
+    apply( proof, { pr => isACNFTop( pr ) },
+      { ( p, cut ) =>
+        cut match {
+          case CutRule( leftSubProof, l, rightSubProof, r ) =>
+            if ( isACNF( p ) )
+              ( ( !InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) || ( InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && !InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) || ( !InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && !InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) )
```
shetzl: This line has a length of 426 characters. Please format it in a way that is more accessible to a human reader by splitting it up into shorter lines.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
+      val p = cutElim( pr, ( isACNF( pr ) && Top ) )( pred )
+      pr = if ( CleanStructRules ) cleanStructuralRules( p ) else p
       if ( recordSteps ) steps += pr
-    } while ( !pred_done( pr ) && !isCutFree( pr ) )
+    } while ( !pred_done( pr ) )
     if ( !recordSteps ) steps += pr
     pr
   }
 
+  def elimToACNFTop( proof: LKProof, ctx: Context, CleanStructRules: Boolean = true ): LKProof =
```
shetzl: Please try to be as consistent as possible when naming things. For this function I would suggest "eliminateToACNFTopByUppermost".

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
+        cut match {
+          case CutRule( leftSubProof, l, rightSubProof, r ) =>
+            if ( isACNF( p ) )
+              ( ( !InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) || ( InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && !InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) || ( !InitOrCut( leftSubProof, leftSubProof.endSequent.apply( l ) ) && !InitOrCut( rightSubProof, rightSubProof.endSequent.apply( r ) ) ) )
+            else !isAtom( leftSubProof.endSequent.apply( l ) ) && !InitAx( leftSubProof ) && !InitAx( rightSubProof ) && isACNF( leftSubProof ) && isACNF( rightSubProof )
+
+        }
+      }, ctx, true, true, CleanStructRules )
+
+  def elimToACNF( proof: LKProof, ctx: Context, CleanStructRules: Boolean = true ): LKProof =
```
shetzl: Analogously to above, here I would suggest "eliminateToACNFByUppermost".

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -124,75 +185,76 @@ class ReductiveCutElimination {
...
     case EqualityRightRule( subProof, _, _, _ ) =>
-      EqualityRightRule( cutElim( subProof ), proof.auxFormulas.head( 0 ), proof.auxFormulas.head( 1 ), proof.mainFormulas.head )
+      EqualityRightRule( cutElim( subProof, Top ), proof.auxFormulas.head( 0 ), proof.auxFormulas.head( 1 ), proof.mainFormulas.head )
 
     case CutRule( leftSubProof, aux1, rightSubProof, aux2 ) =>
       if ( pred( proof ) )
-        reduceGrade( leftSubProof, aux1, rightSubProof, aux2 )
-      else
-        CutRule( cutElim( leftSubProof ), cutElim( rightSubProof ), proof.auxFormulas.head.head )
+        if ( Top ) reduceRankLeft( leftSubProof, aux1, rightSubProof, aux2, Top )
```
shetzl: Why do we have to make a case distinction on the value of Top here? What is wrong with a grade reduction if we are computing an ACNFTop? (This may be easier to discuss in person)

Ermine516:
Actually it is easy to explain. Grade reduction includes atomic cut
elimination. Being that all the cuts are atomic already there is no need
for grade reduction anyway. So I thought it best to just skip it. Note Top
is defined as isacnf (pr) and Acnftop-flag

Am 26.01.2017 12:54 schrieb "Stefan Hetzl" <notifications@github.com>:

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
+             ACNF: Boolean = false, Top: Boolean = false, CleanStructRules: Boolean = true ): LKProof = {
     steps += proof
-
     // For rank-reduction of strong quantifier inferences, we need to either:
     // 1) rename eigenvariables during rank-reduction, or
     // 2) maintain the invariant that the proof is regular
     // Here, we choose 2).
-    var pr = regularize( proof )
+
+    var pr = if ( ctx != null ) eliminateDefinitions( regularize( proof ) )( ctx ) else regularize( proof )
```
shetzl: As discussed, this should not go into master.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: Actually it is easy to explain. Grade reduction includes atomic cut elimination rules. Being that all the cuts are atomic already there is no need for grade reduction anyway. This is checked when cutelim is called. So I thought it best to just skip it.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: I'll make the name changes and summit with comments.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: :) yeah, good point. I'll will format it.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: That was not suppose to be committed. It was for testing.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: I'll remove it based on the discussion yesterday

shetzl:
@Ermine516  This pull request is a good first step. Pushing corrections to the ReducRework branch should automatically update it.

Ermine516:
Putting definition elimination in the method was a quick fix that came out of a discussion with Gabriel because the definition handling mechanism is broken in reductive cut elimination.   We discussed it yesterday that in the end that mechanism should be repaired instead. Didn't remove before committing. Is there really any danger of using null when one is only checking if something has been set? I thought option would be a bit overkill.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
gebner: To be perfectly honest, I don't think I could understand it even if it were split up on multiple lines.  If I understand the definition of ACNFtop correctly, the logic here could basically just be `cut is not in ACNFtop && each sub-proof is cut-free`, right?

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
gebner: If this is just meant as a performance optimization, then we should remove it.  The `reduceGrade` function directly falls through to `reduceRankLeft` if the cut is atomic (and not below a logical axioms), i.e. it is just O(1) overhead.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -311,17 +376,21 @@ class ReductiveCutElimination {
...
+            case ( Seq( aux1Sub ), Seq() ) => // The left cut formula is in the left subproof of the binary inference
+              val cutSub = CutRule( leftSubProof, aux1Sub, right, aux2 )
+              CutRule( cutSub, cutSub.getLeftOccConnector.child( a1 ), rightSubProof, a2 )
+
+            case ( Seq(), Seq( aux1Sub ) ) => // The left cut formula is in the right subproof of the binary inference
+              val cutSub = CutRule( rightSubProof, aux1Sub, right, aux2 )
+              CutRule( leftSubProof, a1, cutSub, cutSub.getLeftOccConnector.child( a2 ) )
+          }
+        } else {
+          reduceRankRight( left, aux1, right, aux2 )
```
gebner: Is this branch ever taken?  As the comment explains, this never happens when eliminating topmost cuts first.  The extra complexity in the !top case is not necessary if we never use it.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -44,60 +48,117 @@ object ReductiveCutElimination {
...
+   * default value is on, i.e. clean the structural rules
+   * @param ACNF Tells the algorithm whether or not to eliminate atomic cuts. Default value
+   *  is on, i.e. eliminate atomic cuts.
+   *
    *
    * @return The proof as it is after pred_done returns true.
    */
-  def apply( proof: LKProof, pred_done: LKProof => Boolean, pred_cut: ( LKProof, LKProof ) => Boolean ): LKProof = {
-
+  def apply( proof: LKProof, pred_done: LKProof => Boolean, pred_cut: ( LKProof, LKProof ) => Boolean, ctx: Context,
```
Ermine516: Yes, you are right, ACNF is not necessary but top is. It is used to skip grade reduction

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: It isn't convenience only though. Grade reduction will try to eliminate the cut so I skip over it.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: yes it is taken. if going to ACNF top without this branch you run into an infinite loop. The cuts keep passing over each other. But this only occurs if the Right branch is reducible.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
Ermine516: no it is more complex then that. The sub proofs have three possible cases. They are both ACNF_top and the last rule on each side is not a cut, a logical axiom, or a weakening of a cut ancestor, there the cut can be ranked reduced on both sides. Or, the other two cases, the left or right branch has as the last rule a rule which is not rule on each side is not a cut, a logical axiom, or a weakening of a cut ancestor,  the cut can be ranked reduced only on that side. It is a tricky algorithm :)

shetzl:
Let me try to re-think the ACNF and ACNFtop computation from scratch.

As with every rewriting system also for cut-elimination there are two degrees of freedom when picking a reduction:

1. which cut to reduce

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -29,7 +30,10 @@ object ReductiveCutElimination {
    * @param proof The proof to subject to cut-elimination.
    * @return The cut-free proof.
    */
-  def apply( proof: LKProof ) = new ReductiveCutElimination().eliminateAllByUppermost( proof )
+  def apply( proof: LKProof, ctx: Context = null, ACNF: Boolean = false, Top: Boolean = false, CleanStructRules: Boolean = true ) =
+    if ( ACNF && Top ) new ReductiveCutElimination().elimToACNFTop( proof, ctx, CleanStructRules )
+    else if ( ACNF ) new ReductiveCutElimination().elimToACNF( proof, ctx, CleanStructRules )
+    else new ReductiveCutElimination().eliminateAllByUppermost( proof, ctx, CleanStructRules )
```
quicquid: Also ```new ReductiveCutElimination().xxx``` should be ```ReductiveCutElimination.xxx``` since this is an object,.

Ermine516:
@quicquid I don't understand what you mean. It was originally the way I have it. I just copied and pasted it. Seems like having new was fine before

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
gebner: @quicquid We have `ReductiveCutElimination` both as class and object.  The original reason was that it has a mutable variable which records the intermediate steps during cut-introduction.  So `new ReductiveCutElimination()` is correct here.

Ermine516: Thanks for the info. Now I know why the code I copied was there :)

quicquid:
@Ermine516 yes, i just saw that it was changed to have mutable variables -- the new is necessary now.

Ermine516:
I like stefan's idea concern the additional predicate. I'll change the code accordingly.  It will get rid of all my weird if statements. One last thing  concerning efficiency of the method. After every redution step it loops through the entire proof looking for a cut, but in we already have a good idea of a cut to reduce from the previous reduction step.  Maybe adding a buffer of plausibly reducible cuts might be a good idea. A cut stays in the buffer till its eliminated or no longer satisfies the pred_cut.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -18,10 +19,21 @@ object ReductiveCutElimination {
...
-   * @return The cut-free proof.
+   * @param proof The proof to subject to cut-elimination or restructuring.
+   * @param mode Tells the algorithm what to output. 0 or None means cut-free proof, 1 means first ACNF found
+   *  2 means ACNF top
+   *  @param cleanStructRules Tells the algorithm whether or not to clean the structural rules
+   * default value is on, i.e. clean the structural rules
+   * @return A proof.
    */
-  def apply( proof: LKProof ) = new ReductiveCutElimination().eliminateAllByUppermost( proof )
+  def apply( proof: LKProof, mode: Option[Int] = None, cleanStructRules: Boolean = true ) = mode match {
```
gebner: Well, switching to integers gets rid of the meaningless `acnf = false, top = true` combination, but introduces countably infinitely many others. :smile:

I think we could make the code here both shorter and more readable by introducing an extra type for reduction strategies that combines the three predicates defined below:
```scala
trait Strategy {
  def isDone(wholeProof: LKProof): Boolean
  def reduce(cut: CutRule): Option[LKProof]
}
```
Then we could have for example
```scala
case object allByUppermost extends Strategy {
  def isDone(wholeProof: LKProof) =
    isCutFree(wholeProof)

  def reduce(cut: CutRule) =
    reduceGrade(cut).
      orElse(reduceRankLeft(cut)).
      orElse(reduceRankRight(cut))
}
```
We would have to move the `reduceGrade` and `reduceRank*` into the companion object, and refactor them to return `Option[LKProof]`.  Then the `apply` method could take a strategy object that defaults to `allByUppermost`, and we could pass it around as an implicit parameter in the `cutElim` method.  What do you think?

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -36,77 +48,167 @@ object ReductiveCutElimination {
...
   }
+
+  /**
+   * This method checks whether a proof is ACNF
+   *
+   * @param proof The proof to check for ACNF.
+   * @return True if proof is ACNF  False otherwise.
+   */
+  def isACNF( proof: LKProof ): Boolean = proof match {
+    case InitialSequent( _ )         => true
```
gebner: It's unreliable to match on nullary/unary/binary LK proofs.  We have LK proofs with 3 sub-proofs and more as well.  The easiest fix here is to remove the cases for nullary/unary/binary rules and just use the following instead (as the last case!):
```scala
  case _ => proof.immediateSubProofs.forall(isACNF)
```

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -36,77 +48,167 @@ object ReductiveCutElimination {
...
+    case UnaryLKProof( _, subProof ) => isACNF( subProof )
+    case CutRule( lsb, l, rsb, _ ) =>
+      if ( isAtom( lsb.endSequent.apply( l ) ) ) isACNF( lsb ) && isACNF( rsb )
```
gebner: BTW, you can also write `lsb.endSequent(l)`.  The `apply` name is special in Scala: if an object has an apply method, then you can call the object like a function (and the apply method is called).

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
shetzl: I agree that `Option[int]` is not a good choice. For the time being, a simpler solution may just be to keep this apply method as it was and to add two more methods `reduceToACNF` and `reduceToACNFTop` to the `ReductiveCutElimination` object.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
shetzl: When I said "please don't delete scaladoc comments" earlier I meant to say "please put it back in" :wink:

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -116,77 +250,75 @@ class ReductiveCutElimination {
...
-      if ( pred( proof ) )
-        reduceGrade( leftSubProof, aux1, rightSubProof, aux2 )
-      else
-        CutRule( cutElim( leftSubProof ), cutElim( rightSubProof ), proof.auxFormulas.head.head )
-  }
+      if ( pred( proof ) ) reduce( proof )
+      else CutRule( cutElim( leftSubProof, reduce ), leftSubProof.endSequent( aux1 ), cutElim( rightSubProof, reduce ), rightSubProof.endSequent( aux2 ) )
 
+  }
+  //reduceGrade( leftSubProof, aux1, rightSubProof, aux2 )
```
gebner: Here is a leftover comment.

core/src/main/scala/at/logic/gapt/proofs/sequents.scala
```diff
@@ -75,6 +75,18 @@ case class Suc( k: Int ) extends SequentIndex {
   def toInt = k
 }
 
+case class MultisetSequent[+A]( sequent: Sequent[A] ) {
+  override def equals( that: Any ): Boolean = that match {
+    case Sequent( ante, suc ) => {
```
gebner: The `that` parameter will be a `MultisetSequent`, not a `Sequent`.  You need to match on `MultisetSequent` here.

core/src/main/scala/at/logic/gapt/proofs/sequents.scala
```diff
@@ -75,6 +75,18 @@ case class Suc( k: Int ) extends SequentIndex {
...
+      ( ( this.sequent.antecedent.distinct.toSet diff ante.distinct.toSet ).isEmpty &&
+        ( this.sequent.succedent.distinct.toSet diff suc.distinct.toSet ).isEmpty )
```
gebner: With this equals method, you should probably call the wrapper `SetSequent`. :smile: 

1. The `diff` method compute the difference, and not the symmetric difference.  (So this method checks the subset relation.)
2. The combination `distinct.toSet` is redundant, as `toSet` already ignores duplicates (it returns a set).
3. You can directly check sets for equality: `sequent.antecedent.toSet == ante.toSet`.

Ermine516: Should I really make a setsequent class?!

gebner: You should name the class after what it does.  If it checks for set-equality, then you should call it `SetSequent`.  If you call it `MultisetSequent`, then it should check for multiset-equality.

Ermine516: I had the impression that I should make another class. Yep now I understand

Ermine516:
Before adding to the main branch I want to fix one thing. Weakening cannot
be ignored and must be pushed to the leaves. Almost done with this fix.

Am 27.01.2017 16:16 schrieb "Stefan Hetzl" <notifications@github.com>:

*@shetzl* commented on this pull request.

The cut-elimination code looks ok to merge for me now. I did not look at
the clause set extraction code.

In the future, I create a new pull request for each feature (in this case:
a new one for clause set extraction).

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/gapt/gapt/pull/600#pullrequestreview-18858716>, or mute
the thread

gebner:
@Ermine516 You just asked me on gitter if there's anything missing in this PR.  I see mainly two things:

 * `MultisetSequent` either needs to be renamed to `SetSequent`, or changed to check multiset equality.
 * Unit tests for the new functionality would be nice.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -63,29 +132,33 @@ class ReductiveCutElimination {
    * left-uppermost cuts until the proof is cut-free) is provided by another
    * apply method in this class.
    *
+   *
+   *
```
shetzl: Please avoid random spacing like this.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -97,17 +170,254 @@ class ReductiveCutElimination {
...
+      cleanStructRules )
+  }
+
+  /**
+   * pushes all weakening rules to the leaves of the proof tree and adds contractions to duplications
+   *
+   * @param proof An LKProof.
+   * @return A proof weakening rules at the leaves only.
+   */
+  def pushWeakeningToLeaves( proof: LKProof ): LKProof = proof match {
```
shetzl: As discussed in person a while ago, pushing weakening to the top should not be part of cut-elimination. It's a proof transformation of independent interest.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -97,17 +170,254 @@ class ReductiveCutElimination {
...
+    case CutRule( leftSubProof, aux1, rightSubProof, aux2 ) =>
+      CutRule( pushWeakeningToLeaves( leftSubProof ), leftSubProof.endSequent( aux1 ), pushWeakeningToLeaves( rightSubProof ), rightSubProof.endSequent( aux2 ) )
+
+  }
+
+  /**
+   * pushes all weakening rules to the leaves of the proof tree and adds contractions to duplications
+   *
+   * @param proof An LKProof.
+   * @param side  False means weakening left true means weakening right
```
shetzl: We have a class Polarity that can be used instead of a Boolean for `side`.

tests/src/test/scala/at/logic/gapt/proofs/lk/AtomicCutNormalFormViaReductiveCutEliminationTest.scala
```diff
@@ -0,0 +1,199 @@
+package at.logic.gapt.proofs.lk
```
shetzl: These tests should go into `ReductiveCutEliminationTest.scala` to be consistent with the code in src.

Ermine516:
I thought it was finished. We can discuss today what to do to finish it

Am 01.03.2017 11:31 schrieb "Stefan Hetzl" <notifications@github.com>:

core/src/main/scala/at/logic/gapt/proofs/lk/lk.scala
```diff
@@ -65,6 +65,16 @@ abstract class LKProof extends SequentProof[HOLFormula, LKProof] {
  *     Γ' :- Δ'
  * </pre>
  */
+
```
gebner: Duplicate scaladoc comment here.

examples/Schema/NiaSchema.scala
```diff
@@ -0,0 +1,271 @@
+package at.logic.gapt.examples
```
gebner: Schema code.  Does not even compile.

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -116,86 +247,83 @@ class ReductiveCutElimination {
...
   /**
-   * Grade reduction rules, i.e. rules that reduce the complexity of a cut formula or remove a cut altogether.
-   *
-   * @param left The left subproof of the cut inference.
-   * @param aux1 The index of the cut formula in the left subproof.
-   * @param right The right subproof of the cut inference.
-   * @param aux2 The index of the cut formula in the right subproof.
-   * @return
-   */
+    * Grade reduction rules, i.e. rules that reduce the complexity of a cut formula or remove a cut altogether.
```
gebner: Can you please run `sbt test` before comitting, every time?  I get the feeling that this comment would be reformatted by scalariform (when you run sbt).

core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
gebner: Extra spacing still there.

PR #601, opened 2017-01-27
new CERES method 

anelaLogic:
new CERES method which extracts expansion proofs from projections and the resolution refutation,
changes in ResolutionToExpansionProof to handle cases where we call the method with the new CERES method
new testProof in simple

### DIFF
--- a/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
@@ -9,7 +9,7 @@ import at.logic.gapt.provers.sat.Sat4j
 import scala.collection.mutable
 
 /**
- * Converts a resolution proof to an expansion proof.*
+ * Converts a resolution proof to an expansion proof.
  * ResolutionToExpansionProof( rp ) will return the expansion proof of rp and ResolutionToExpansionProof( rp, input )
  * is used by the CERES method to return the expansion proof of the ACNF, which is extracted from the proof projections.
  *
@@ -62,13 +62,13 @@ object ResolutionToExpansionProof {
         Sequent() :+ ( ETMerge( f, Polarity.InSuccedent, set.map( _._2.elements.head ) ) )
 
       case ( Input( Sequent( Seq(), Seq( f ) ) ) ) if freeVariables( f ).isEmpty =>
-        Sequent().+:( ETMerge( f, Polarity.InAntecedent, set.map( _._2.elements.head ) ) )
+        ETMerge( f, Polarity.InAntecedent, set.map( _._2.elements.head ) ) +: Sequent()
 
       case ( Input( seq ) ) =>
         val fvs = freeVariables( seq ).toSeq
         val sh = All.Block( fvs, seq.toDisjunction )
-        Sequent().+:( ETWeakQuantifierBlock( sh, fvs.size,
-          for ( ( subst, es ) <- set ) yield subst( fvs ) -> es.toDisjunction( Polarity.Negative ) ) )
+        ETWeakQuantifierBlock( sh, fvs.size,
+          for ( ( subst, es ) <- set ) yield subst( fvs ) -> es.toDisjunction( Polarity.Negative ) ) +: Sequent()
     }
   }
 
--- a/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
@@ -37,7 +37,7 @@ import scala.collection.mutable
  *
  * In the unlikely case that the resolution proof starts from clauses, we just pretend that the clauses are derived
  * from formulas--that is, we instead of beginning with `D(x), E(x,y) :- F(y)` we just imagine the proof begins with
- * `:- ∀x ∀y (¬D(x) ∨ ¬E(x,y) ∨ F(y))`.*
+ * `:- ∀x ∀y (¬D(x) ∨ ¬E(x,y) ∨ F(y))`.
  *
  * Splitting inferences are converted to cuts in the expansion proof: formally, we can first skip all the splitting
  * inferences in the resoltion proof; the assertions now become part of the clauses, and we get additional input sequents.
--- a/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
@@ -9,7 +9,9 @@ import at.logic.gapt.provers.sat.Sat4j
 import scala.collection.mutable
 
 /**
- * Converts a resolution proof to an expansion proof.
+ * Converts a resolution proof to an expansion proof.*
+ * ResolutionToExpansionProof( rp ) will return the expansion proof of rp and ResolutionToExpansionProof( rp, input )
+ * is used by the CERES method to return the expansion proof of the ACNF, which is extracted from the proof projections.
  *
  * Let us first ignore splitting and subformula definitions for simplicity.
  * However we may still have clausification inferences.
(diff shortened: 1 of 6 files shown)


shetzl:
As the travis build shows, there is the following test failure:
```
[error] /home/travis/build/gapt/gapt/tests/src/test/scala/at/logic/gapt/proofs/resolution/ResolutionTest.scala:114: not enough arguments for method withDefs: (proof: at.logic.gapt.proofs.resolution.ResolutionProof, input: (at.logic.gapt.proofs.resolution.Input, at.logic.gapt.proofs.expansion.ExpansionSequent, Set[(at.logic.gapt.expr.Substitution, at.logic.gapt.proofs.expansion.ExpansionSequent)]) => at.logic.gapt.proofs.expansion.ExpansionSequent, addConclusion: Boolean)at.logic.gapt.proofs.expansion.ExpansionProofWithCut.
[error] Unspecified value parameter input.
[error]     ResolutionToExpansionProof.withDefs( proof ).deep must beValidSequent
[error]                                        ^
[error] one error found
[error] (tests/test:compileIncremental) Compilation failed
[error] Total time: 21 s, completed Jan 27, 2017 12:12:25 PM
```

Please make sure that `sbt test` finishes sucessfully.

tests/src/test/scala/at/logic/gapt/proofs/resolution/ResolutionTest.scala
```diff
@@ -111,7 +112,26 @@ class ResolutionTest extends Specification with SatMatchers {
     val proof = Resolution( Resolution( AvatarContradiction( split ), Suc( 0 ), case1, Ant( 0 ) ), Suc( 0 ), case2, Ant( 0 ) )
     proof.isProof must_== true
 
-    ResolutionToExpansionProof.withDefs( proof ).deep must beValidSequent
+    var retSeq: ExpansionSequent = Sequent()
+    ResolutionToExpansionProof.withDefs( proof, {
```
shetzl: Please don't copy&paste code. I'm sure there is a better solution...

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
```diff
@@ -51,7 +53,31 @@ import scala.collection.mutable
 object ResolutionToExpansionProof {
 
   def apply( proof: ResolutionProof ): ExpansionProof = {
-    val expansionWithDefs = withDefs( proof )
+    var retSeq: ExpansionSequent = Sequent()
```
gebner: It took me a few minutes to understand how `retSeq` is modified here.  I think we can considerably simplify this code if we drop the `ExpansionSequent` argument in the function passed to apply, and instead let `withDefs` do the concatenation.

That is, we should change the type of `apply` to:
```scala
def apply(
    proof: ResolutionProof,
    input: ( Input, Set[( Substitution, ExpansionSequent )] ) => ExpansionSequent
  ): ExpansionProof
```
And `withDefs` would just do:
```scala
  expansionSequent ++= input( Input( seq ), expansionSequent, expansions( p ) )
```

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
```diff
@@ -122,20 +148,31 @@ object ResolutionToExpansionProof {
     def sequent2expansions( sequent: HOLSequent ): Set[( Substitution, ExpansionSequent )] =
       Set( Substitution() -> sequent.zipWithIndex.map { case ( a, i ) => ETAtom( a.asInstanceOf[HOLAtom], !i.polarity ) } )
 
+    def perfMerges( expansionSequent: ExpansionSequent ): ExpansionSequent = {
+      var expSeq = expansionSequent
+      var retSeq: ExpansionSequent = Sequent()
```
gebner: In general it is a good idea to prefer immutable over mutable variables, and existing functions on collections (i.e. `map`, `flatMap`, `sum`, etc.).  It usually makes the code shorter and easier to comprehend for other people.

Using `groupBy` and `map` this function is just one line, and moreover it is clear what it does: it groups the expansion trees on each side by shallow formula, and then merges each group:
```scala
expansionSequent.groupBy(_.shallow).map(ets => ETMerge(ets._2))
```

(The asymptotic complexity is also much better: the short version is linear*, while the long one is quadratic.  *Linear insofar as hash table insertion is O(1).)

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
anelaLogic: That was my first intention, too. However, I could not manage to transform an ExpansionTree (which is the result of ETMerge) into an ExpansionSequent. This is what we would need to do in the case you call apply without an input function and have to return an ExpansionSequent.

gebner: This creates an expansion sequent with one formula in the succedent:
```scala
Sequent() :+ ETMerge(f, Polarity.InSuccedent,
                     set.map(_._2.elements.head))
```

anelaLogic: ok, then it's easy

tests/src/test/scala/at/logic/gapt/proofs/resolution/ResolutionTest.scala
```diff
@@ -113,25 +113,7 @@ class ResolutionTest extends Specification with SatMatchers {
     proof.isProof must_== true
 
     var retSeq: ExpansionSequent = Sequent()
```
gebner: This variable is now unused.

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
```diff
@@ -53,30 +53,26 @@ import scala.collection.mutable
...
-    var retSeq: ExpansionSequent = Sequent()
-    val exp = apply( proof, {
-      case ( Input( Sequent( Seq( f ), Seq() ) ), expSeq, set ) if freeVariables( f ).isEmpty =>
-        retSeq = expSeq
-        retSeq :+= ETMerge( f, Polarity.InSuccedent, set.map( _._2.elements.head ) )
-        retSeq
+    apply( proof, setInput() )
+  }
+
+  def setInput()( input: Input, set: Set[( Substitution, ExpansionSequent )] ): ExpansionSequent = {
```
gebner: Minor nitpick: you don't need the first `()` here.

Do you have an idea for a more meaningful function name than `setInput` here?  If I see `setInput` somewhere, I can't really tell what it's supposed to do.  Maybe `inputsAsExpansionSequent`?

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
```diff
@@ -53,30 +53,26 @@ import scala.collection.mutable
...
+    input match {
+      case ( Input( Sequent( Seq( f ), Seq() ) ) ) if freeVariables( f ).isEmpty =>
+        Sequent().:+( ETMerge( f, Polarity.InSuccedent, set.map( _._2.elements.head ) ) )
```
gebner: In Scala there is no real distinction between infix operators and methods, instead of `a * b + c`, you can always write `a.*(b.+(c))` as you've discovered here.
We usually use the infix notation `sequent :+ formula` for these methods though.

core/src/main/scala/at/logic/gapt/proofs/resolution/expansionProofFromInstances.scala
```diff
@@ -19,12 +19,13 @@ object expansionProofFromInstances {
       subst <- substs
     } yield Subst( just, subst )
 
+    var retSeq: ExpansionSequent = Sequent()
```
gebner: This variable is now unused as well.

core/src/main/scala/at/logic/gapt/proofs/resolution/ResolutionToExpansionProof.scala
anelaLogic: it was a bit confusing, as I could not use the infix notation for " :+" but for "+:"
Now I changed "+:" to infix.

gebner: That's because `+:` goes "on the other side" (since it ends with a colon):
```scala
  a +: Sequent()    ==   Sequent().+:(a)    // add on the left
  Sequent() :+ a    ==   Sequent().:+(a)    // add on the right
```

PR #604, opened 2017-02-02
Pi2 cut intro.git

Lettmann:


core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
```diff
@@ -0,0 +1,266 @@
...
+    val ( posAtoms, negAtoms ) = literals
+
+    val unifiedLiterals = scala.collection.mutable.Set[FOLFormula]()
+
+    posAtoms.foreach( posAt =>
+      negAtoms.foreach( negAt =>
+        unifyLiterals(
+          posAt,
+          negAt,
+          productionRulesXS.toSet,
```
gebner: You have multiple functions that have the same set of unchanged parameters.  Passing them from one function to another just adds unnecessary "line noise" that hides the actual code.  You have multiple options to clear this up:
1) Nest the functions inside a big one:
```scala
def bigFunction(productionRulesXS: Set[...], productionRulesYS: Set[...], ...) = {
  def unify(posAt: ..., negAt: ...) = {
    // you can now use productionRulesXS here as well
  }

  def unifyLiterals(...) = {
    ...
  }
}
```
2) Put all of the functions inside a class:
```scala
class GStarUnifier(productionRulesXS: Set[...], productionRulesYS: Set[...], ...) = {
  def unify(posAt: ..., negAt: ...) = {
    // you can use productionRulesXS here
  }

  def unifyLiterals(...) = {
    ...
  }
}
```
3) Combine the arguments into a single structure, then you only need to pass a single argument:
```scala
case class Pi2Grammar(productionRulesXS: Set[...], productionRulesYS: Set[...], ...)

object gStarUnify {
  def unify(posAt: ..., negAt: ..., grammar: Pi2Grammar) = {
    ...
  }

  def unifyLiterals(..., grammar: Pi2Grammar) = {
    ...
  }
}
```

I think option 3 is a good choice here.

core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
```diff
@@ -0,0 +1,266 @@
...
+    zippedArgs:                List[( LambdaExpression, LambdaExpression )],
+    productionRulesX:          Set[( LambdaExpression, LambdaExpression )],
+    productionRulesY:          Set[( LambdaExpression, LambdaExpression )],
+    universalEigenvariable:    FOLVar,
+    existentialEigenvariables: List[FOLVar],
+    nameOfExistentialVariable: FOLVar,
+    nameOfUniversalVariable:   FOLVar
+  ): Option[Seq[FOLTerm]] = {
+
+    var unifiedTerms: Option[Seq[FOLTerm]] = None
```
gebner: Is there a difference between `Some(Set())` and `None`?

core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
```diff
@@ -0,0 +1,266 @@
...
+      stopIt = false
+      stopItAll = false
+
+      val ( tL, tR ) = t
+
+      productionRulesX.foreach( productionRuleX => if ( !stopIt ) {
+
+        val ( productionRuleXL, productionRuleXR ) = productionRuleX
+
+        if ( productionRuleXL.syntaxEquals( tL ) && productionRuleXR.syntaxEquals( tR ) ) {
```
gebner: Use `==` instead of `syntaxEquals`.

core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
```diff
@@ -0,0 +1,266 @@
...
+        theUnifiedLiteral
+      }
+      case _ => None
+    }
+
+    unifiedLiteral
+
+  }
+
+  private def unify(
```
gebner: This function is terribly complex, and I wouldn't be surprised if there were a few bugs hiding here.  Unfortunately I failed to understand what it should do.

A few suggestions:
 * The functions takes 1) a look at each argument pair, and then 2) decomposes the argument pairs again into function and argument parts.  It think it's a good idea to move part 2) into a separate function, then each function is of a more manageable size.  Is part 2) maybe even similar to part 1), and could they be combined?
 * Try to get rid of the mutable flags and variables.

core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
```diff
@@ -0,0 +1,266 @@
...
+            unifiedTerms = unifiedTerms match {
+              case Some( update ) => Option( update :+ nameOfExistentialVariable )
+              case None           => Option( Seq( nameOfExistentialVariable ) )
+            }
+
+          } )
+
+          if ( !tLWasAEigenvariable && ( argsOfArgL.length == 0 ) ) {
+
+            tLWasAEigenvariable = true
```
gebner: Why is `tL` an eigenvariable now?   Or was?  Does it change??

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -0,0 +1,680 @@
...
+    val existentialEigenvariables:     List[FOLVar], // beta_1,...,beta_m
+    val substitutionsForAlpha:         List[LambdaExpression], // r_1,...,r_m
+    val substitutionsForBetaWithAlpha: List[LambdaExpression] // t_1(alpha),...,t_p(alpha)
+) {
+
+  require( existentialEigenvariables.length == substitutionsForAlpha.length )
+
+  val multiplicityOfAlpha: Int = substitutionsForAlpha.length // m
+  val multiplicityOfBeta: Int = substitutionsForBetaWithAlpha.length // p
+  var balancedSolution: Option[FOLFormula] = None
```
gebner: Don't put mutable fields into data structures.

In this case, these two fields just store the result of another function, make the other function return these!

(I'm not sure, but is `noSolutionHasBeenFound` redundant?  Does it mean something different from `balancedSolution == None`?)

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -0,0 +1,680 @@
...
+  var noSolutionHasBeenFound: Boolean = true
+
+  // (alpha,r_1),...,(alpha,r_m)
+  //////////////////////////////
+  def substitutionPairsAlpha(): Set[( LambdaExpression, LambdaExpression )] = {
+
+    val substitutionPairsAlpha = scala.collection.mutable.Set[( LambdaExpression, LambdaExpression )]()
```
gebner: Take a look at the `map` method of Set.

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -0,0 +1,680 @@
+package at.logic.gapt.expr
+
+import at.logic.gapt.expr.hol.CNFp
+import at.logic.gapt.proofs.Sequent
+
+/**
+ * Created by root on 26.01.17.
+ */
+class pi2SeHs(
```
gebner: It is idiomatic to make data structures case classes in Scala:
```scala
case class Pi2SeHS(
  reducedRepresentation: Sequent[FOLFormula],
  ...
)
```

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -0,0 +1,680 @@
...
+  }
+
+  // (beta_1,t_1(alpha)),...,(beta_1,t_p(alpha)),
+  //                     ...                    ,
+  // (beta_m,t_1(alpha)),...,(beta_m,t_p(alpha))
+  ///////////////////////////////////////////////
+  def substitutionPairsBeta(): Set[( LambdaExpression, LambdaExpression )] = {
+
+    val substitutionPairsBeta = scala.collection.mutable.Set[( LambdaExpression, LambdaExpression )]()
+    for ( index <- 1 to multiplicityOfAlpha ) {
```
gebner: Check out the following:
```scala
(
  for (index <- 1 to multiplicityAlpha)
    yield substitutionPairsBetaI(multiplicityOfAlpha - index + 1)
).toSet
```

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -0,0 +1,680 @@
...
+
+    resultingSequent
+  }
+
+  private def substituteLeftSideOnce( sequent: Sequent[HOLFormula], index: Int ): Sequent[HOLFormula] = {
+
+    var resultingSequent: Sequent[HOLFormula] = Sequent()
+
+    sequent.antecedent.foreach( formula => {
+      formula.find( existentialEigenvariables( index - 1 ) ) match {
```
gebner: It took me a second to understand this, I think `freeVariables(formula).contains(existentialEigenvariables(index-1))` is clearer for the condition.

However, do you even need to handle that case specially?  If you just want to get rid of duplicate formulas, you can call `.distinct` on the sequent later on.

core/src/main/scala/at/logic/gapt/expr/gStarUnify.scala
Lettmann: I used == at the beginning, but got some mistakes. So I changed it and now it is working.

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -193,6 +193,79 @@ class pi2SeHs(
...
+    val ( _, oldDNTAs ) = this.literalsInTheDNTAsAndTheDNTAs
+    val newDNTAs = oldDNTAs.map( leaf => {
+      leaf.antecedent.filter( literal => {
+        literal match {
+          case Neg(t) => {
+            val isNotInLanguage: Boolean = unifiedLiterals.forall( opponent => {
+              val Apps(nameOfLiteral,_) = t
+              val Apps(nameOfOpponent,_) = opponent
+              if (nameOfLiteral.syntaxEquals(nameOfOpponent)) {
+                false
```
gebner: This is a novel way to write `nameOfLiteral != nameOfOpponent`. :smile:

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -193,6 +193,79 @@ class pi2SeHs(
...
+              } else {
+                true
+              }
+            })
+            !isNotInLanguage
+          }
+          case t => {
+            val isNotInLanguage: Boolean = unifiedLiterals.forall( opponent => {
```
gebner: You might want to apply de Morgan's law here, and commute the negation with the `forall`:
```scala
val isInLanguage = unifiedLiterals.exists {
  case Apps(nameOfOpponent, _) => nameOfLiteral == nameOfOpponent
}
```

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -193,6 +193,79 @@ class pi2SeHs(
...
+            if (nameOfLiteral.syntaxEquals(nameOfOpponent)) {
+              false
+            } else {
+              true
+            }
+          })
+            !isNotInLanguage
+          }
+        }
+      }).foldLeft(Sequent[FOLFormula]())( (a,b) => b +: a )
```
gebner: This line should add the formulas on the left side of the sequent, right?  There is already the `++:` method for that:
```scala
... ++: Sequent()
```

core/src/main/scala/at/logic/gapt/expr/introducePiCut.scala
```diff
@@ -332,10 +405,12 @@ class ClausesWithIndexLists(
 
     var bool: Boolean = true
     if ( clauses.length == 1 ) {
-      if ( clauses.head.isAllowedAtLeastAsSubformula ) {
+      if (clauses.head.isAllowedAtLeastAsSubformula) {
         this.leafIndexListClauses.forall( leafNumber => {
           if ( leafNumber.oneToPList.isEmpty ) {
             bool = false
+          } else if ( leafNumber.oneToMList.isEmpty ) {
```
gebner: I think we can drop the `bool` variable here.  If I see it correctly, then this whole block is just a roundabout way to write:
```scala
!leafNumber.oneToPList.isEmpty && !leafNumber.oneToMList.isEmpty
```

tests/src/test/scala/at/logic/gapt/expr/IntroducePiCutTest.scala
```diff
@@ -131,4 +129,92 @@ class IntroducePiCutTest extends Specification {
...
+      val B1 = fof"$Pxf1x->$Pxff1x"
+      val B2 = fof"$Pxf2x->$Pxff2x"
+      val C3 = fof"$Pcfy1&$Pfy1fy2->$Pcgy2"
+      val D = fof"$Pcgy2"
+      val Rere = A2 +: B1 +: B2 +: C3 +: Sequent() :+ D
+      val seHs = new pi2SeHs( Rere, fov"x", List( fov"y1", fov"y2" ), List( fot"c", fot"f(y1)" ), List( fot"f1(x)", fot"f2(x)" ) )
+      val xName = fov"xName"
+      val yName = fov"yName"
+      introducePi2Cut( seHs, yName, xName ) must beOneOf (
+        Option( fof"P($xName)&Q(f($yName))|P(f($yName))&Q($xName)" ),
```
gebner: You should write `Some` instead of `Option`.  The `Option` constructor is only intended for interoperability with Java code, the difference is that `Option` will transparently convert `null` to `None`.

PR #607, opened 2017-02-08
Add general definition rules.

gebner:
We can change an auxiliary formula into the main formula if they are definitionally equal.  That is, when they reduce to the same normal form in the current context.

The double-negation example works trivially.

If there are no objections, I'll merge this tomorrow.

### DIFF
--- a/core/src/main/scala/at/logic/gapt/expr/BetaReduction.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/BetaReduction.scala
@@ -38,6 +38,9 @@ class Normalizer( rule: ReductionRule ) {
     Apps( hd_, as_ )
   }
 
+  def isDefEq( a: LambdaExpression, b: LambdaExpression ): Boolean =
+    normalize( a ) == normalize( b )
+
   @tailrec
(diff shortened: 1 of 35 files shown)


shetzl:
Please also update the user manual, in particular the appendix describing the calculi.

gebner:
We can now support a limited class of definition rules during expansion proof extraction.  The only not completely covered case is when an equality rule is below a definition rule.  Then `replaceWithContext` needs to permute the rewriting with the definition rule.  This is implemented using a simple ad-hoc matching algorithm that unfolds definitions (`commuteReplacementCtxWithDefEq`), it requires the context to support the case of equality below definition rules.

Moving definitions upwards in an expansion tree is relatively easy with this approach, but it requires a context as well to work nicely (i.e. to introduce definitions as late as possible).

Unfortunately, none of this is enough yet to extract an expansion proof from the Fürstenberg proof: there we use higher-order rewriting...  The replacement context is `^x x(m_0)`, BTW.

gebner:
@shetzl It seems like we never updated the user manual to describe the current restricted definition rules:

> These definition rules are extremely liberal, as they allow the replacement of any formula by any
other formula.

I'll add a line about the context, though.

PR #608, opened 2017-02-12
Git branch schema

Ermine516:


Ermine516:
btw, it did not work importing niaSchema._ but removing most annotations worked.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -240,6 +241,22 @@ object Context {
...
 
+  case class ProofNames( names: Map[String, ( LambdaExpression, HOLSequent )] ) {
+    def +( name: String, linkpression: LambdaExpression, linkquent: HOLSequent ) = copy( names + ( ( name, ( linkpression, linkquent ) ) ) )
+  }
+
+  implicit val ProofsFacet: Facet[ProofNames] = Facet( ProofNames( Map[String, ( LambdaExpression, HOLSequent )]() ) )
+
+  case class ProofDefinitions( components: Map[String, Set[( LambdaExpression, LKProof )]] ) {
+    def +( name: String, linkpression: LambdaExpression, linkproof: LKProof ) = {
+      components.get( name ) match {
```
gebner: Check out `components.getOrElse(name, Set())`, then you don't need the case split.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+  case class ProofNameDeclaration( lhs: LambdaExpression, endSequent: HOLSequent ) extends Update {
+    override def apply( ctx: Context ): State = {
+      endSequent.foreach( ctx.check( _ ) )
+      val fvEs = freeVariables( endSequent )
+      lhs match {
+        case Apps( at.logic.gapt.expr.Const( c, t ), vs: Seq[LambdaExpression] ) => {
+          val nameIsThere = ctx.get[ProofNames].names.keySet.contains( c )
+          val varcheck: Boolean = vs == vs.distinct &&
+            vs.forall( _.isInstanceOf[Var] ) &&
+            ( freeVariables( endSequent ).toSet[LambdaExpression] ).subsetOf( vs.toSet ) &&
```
gebner: Sets have equality implemented, so you can do just `freeVariables(endSequent) == vs.toSet` here.

From a stylistic point of view, instead of combining all the conditions with a conjunction, I'd write separate require statements here:
```scala
require(!ctx.get[ProofNames].names.contains(c))
require(vs == vs.distinct)
// ...
```

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
         .update[SkolemFunctions]( _ + ( sym, defn ) )
     }
   }
+
+  case class ProofNameDeclaration( lhs: LambdaExpression, endSequent: HOLSequent ) extends Update {
+    override def apply( ctx: Context ): State = {
+      endSequent.foreach( ctx.check( _ ) )
+      val fvEs = freeVariables( endSequent )
+      lhs match {
+        case Apps( at.logic.gapt.expr.Const( c, t ), vs: Seq[LambdaExpression] ) => {
```
gebner: The type-bound on `vs` is unnecessary here.  Also, please avoid using fully-qualified class names and use the simple names instead when possible (that is, just `Const`).

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+          val nameIsThere = ctx.get[ProofNames].names.keySet.contains( c )
```
gebner: You can drop the `.keySet` here, Map has a contains method as well.

core/src/main/scala/at/logic/gapt/proofs/gaptic/tactics/lkTactics.scala
```diff
@@ -1,12 +1,27 @@
...
 import at.logic.gapt.proofs.lk._
 
 /**
+ * Closes a goal with a proof link
+ *
+ * @param proofName The name of the proof proving the goal.
+ */
+case class ProofLinkTactic( proofName: String )( implicit ctx: Context ) extends Tactic[Unit] {
+  def apply( goal: OpenAssumption ) = ctx.get[ProofNames].names.get( proofName ) match {
+    case Some( ( Apps( c, args ), linkEs ) ) => clauseSubsumption( linkEs, goal.conclusion ) match {
```
gebner: You don't need to match on `Apps` here (it will always succeed, BTW).

core/src/main/scala/at/logic/gapt/proofs/lk/LKProofSchemata.scala
```diff
@@ -0,0 +1,180 @@
...
+   * Given a proof name found in the context and a set of arguments matching
+   * the argument list of the chosen proof name this function constructs an
+   * instantiation of that proof. Note that this can result in an infinite
+   * loop or no proof depending on how the chosen arguments are used in the
+   * the chosen proof
+   *
+   * @param ProofName The name of the linkProof
+   * @param args The arguments for the free parameters of the linkProof.
+   */
+  def Instantiate( ProofName: String, args: Seq[LambdaExpression] )( implicit ctx: Context ): Option[LKProof] = {
```
gebner: Please use lower-case camelCase names for methods and arguments.

core/src/main/scala/at/logic/gapt/proofs/lk/LKProofSchemata.scala
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+  /**
+   * This function traverses a schematic proofs and applies the proof and variable substitutions where
+   * needed.
+   *
+   * @param subs The substitutions for the free parameters
+   * @param linkProof The proof we are traversing
+   */
+  def buildProof( subs: Substitution, linkProof: LKProof )( implicit ctx: Context ): Option[LKProof] = linkProof match {
```
gebner: This function should be redone using `LKProofVisitor`.  Also, the substitution should be applied before calling `buildProof`, then we don't need to worry about eigenvariable renaming here.

core/src/main/scala/at/logic/gapt/proofs/lk/LKVisitor.scala
```diff
@@ -42,6 +42,8 @@ trait LKVisitor[T] {
     case p: TheoryAxiom =>
       visitTheoryAxiom( p, otherArg )
 
+    case p: ProofLink =>
+      visitProofLink( p, otherArg )
```
gebner: Nitpick: please add an empty line here, to be consistent with the other cases.

core/src/main/scala/at/logic/gapt/proofs/lk/lk.scala
```diff
@@ -65,6 +65,16 @@ abstract class LKProof extends SequentProof[HOLFormula, LKProof] {
  *     Γ' :- Δ'
  * </pre>
  */
+
+/**
```
gebner: There are two copies of the scaladoc comment here.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+          val varcheck: Boolean = vs == vs.distinct &&
+            vs.forall( _.isInstanceOf[Var] ) &&
+            ( freeVariables( endSequent ).toSet[LambdaExpression] ).subsetOf( vs.toSet ) &&
+            vs.toSet.subsetOf( ( freeVariables( endSequent ).toSet[LambdaExpression] ) ) &&
+            !nameIsThere
+          if ( varcheck ) ctx.state.update[ProofNames]( _ + ( c, lhs, endSequent ) )
+          else throw new IllegalArgumentException( "variables of " + lhs.toString() + "   " + vs.toString() +
```
gebner: Check out string interpolation: `s"variables of $lhs   $vs do not match ..."`

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+    }
+  }
+
+  case class ProofDefinitionDeclaration( lhs: LambdaExpression, linkProof: LKProof ) extends Update {
+    override def apply( ctx: Context ): State = {
+      linkProof.endSequent.foreach( ctx.check( _ ) )
+      lhs match {
+        case Apps( at.logic.gapt.expr.Const( c, t ), vs ) => {
+          vs.foreach( ctx.check( _ ) )
+          val decName = ctx.get[ProofNames].names.keySet.map( key => ctx.get[ProofNames].names.get( key ).get )
```
gebner: The combination `foo.keySet.map(key => foo.get(key).get)` is slightly redundant.  This is just `foo.values`.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+            .fold( None: Option[( LambdaExpression, HOLSequent )] )( ( x, y ) => {
```
gebner: I think the whole check becomes clearer if you just use `exists`:
```scala
require(ctx.get[ProofNames].values.exists { case (name, _) => syntacticMatching(name, lhs).isDefined })
```
We should also check whether end-sequent in the definition matches the declaration.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -354,4 +371,64 @@ object Context {
...
+      }
+
+    }
+  }
+
+  case class ProofDefinitionDeclaration( lhs: LambdaExpression, linkProof: LKProof ) extends Update {
+    override def apply( ctx: Context ): State = {
+      linkProof.endSequent.foreach( ctx.check( _ ) )
+      lhs match {
+        case Apps( at.logic.gapt.expr.Const( c, t ), vs ) => {
```
gebner: If you want to match a single pattern and throw an exception otherwise, you can just write an assignment:
```scala
val Apps(Const(c,t), vs) = lhs
```
This will throw a MatchError if the match fails.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
Ermine516: I always felt getOrElse is a bit to threatening for my taste, but it does clean up the code a bit. :D

core/src/main/scala/at/logic/gapt/proofs/Context.scala
Ermine516: I really should learn that a lot of this stuff exists.

tests/src/test/scala/at/logic/gapt/proofs/lk/SchemaTest.scala
```diff
@@ -0,0 +1,77 @@
...
+  "Testing gNia-schema both parameters non-zero" in {
+    val proof = LKProofSchemata.Instantiate("phi", Seq(le"(s(s(s((s:w>w) (0:w)))))", le"(s(s(s((s:w>w) (0:w)))))"))(gniaSchema.ctx)
+    gniaSchema.ctx.check(proof)  
+
+   None mustEqual None   }
+
+  "Testing gNia-schema both parameters non-zero large" in {
+    val proof = LKProofSchemata.Instantiate("phi",Seq(le"((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) (0:w)))))))))))",le"( ( (((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) ((s:w>w) (0:w))))))))))))))"))(gniaSchema.ctx)
+    gniaSchema.ctx.check(proof)  
+   None mustEqual None   }
```
gebner: Please don't forget to run scalarifom before committing!  (It is automatically run if you execute the tests using sbt.)

core/src/main/scala/at/logic/gapt/proofs/lk/LKProofSchemata.scala
```diff
@@ -0,0 +1,87 @@
...
+    override def visitProofLink( proof: ProofLink, otherArg: Context ): ( LKProof, SequentConnector ) = {
+      val Apps( Const( c, _ ), vs ) = proof.referencedProof
+      val upProof = Instantiate( c, vs )( otherArg )
+      ( upProof, SequentConnector.guessInjection( upProof.endSequent, proof.conclusion ).inv )
+    }
+  }
+
+}
+
+class CarriageReturnList {
```
gebner: What is a carriage-return list?  It does not seem to be used anywhere.

core/src/main/scala/at/logic/gapt/proofs/gaptic/tactics/lkTactics.scala
```diff
@@ -2,11 +2,26 @@ package at.logic.gapt.proofs.gaptic.tactics
...
 /**
+ * Closes a goal with a proof link
+ *
+ * @param proofName The name of the proof proving the goal.
+ */
+case class ProofLinkTactic( proofName: String )( implicit ctx: Context ) extends Tactic[Unit] {
+  def apply( goal: OpenAssumption ) = ctx.get[ProofNames].names.get( proofName ) match {
+    case Some( ( term, linkEs ) ) => clauseSubsumption( linkEs, goal.conclusion ) match {
+      case Some( sub ) => Right( (), ProofLink( sub( term ), sub( linkEs ) ) )
+      case None        => Left( TacticalFailure( this, "Mismatch between  goal " + goal.toString + " and  Linkquent " + linkEs.toString ) )
```
gebner: Here the code still uses the name linkquent, I think we should switch to referenced sequent to be consistent.

core/src/main/scala/at/logic/gapt/proofs/lk/LKProofSchemata.scala
Ermine516: I was implementing this at some point, I will remove it from this pull request.

core/src/main/scala/at/logic/gapt/proofs/gaptic/tactics/lkTactics.scala
Ermine516: Should I make the changes in Context as well.

gebner: Yes, please.

Ermine516: Fixed

PR #611, opened 2017-02-28
Polymorphize all the things

gebner:
Let's add polymorphic definitions, and parametric polymorphism for types!  I think the unit test captures the general idea pretty well:
```scala
      ctx += Context.InductiveType(
        ty"list ?a",
        hoc"nil: list ?a",
        hoc"cons: ?a > list ?a > list ?a"
      )
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,6 +46,7 @@ lazy val commonSettings = Seq(
 val specs2Version = "3.8.8"
 lazy val testSettings = Seq(
   testOptions in Test += Tests.Argument( TestFrameworks.Specs2, "junitxml", "console" ),
+  javaOptions in Test += "-Xmx2g",
   libraryDependencies ++= Seq(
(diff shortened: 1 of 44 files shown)


shetzl:
As discussed on Monday, with this pull request, at the latest, our type theory has reached a level of complexity which merits a thorough explanation in the user manual.

gebner:
@jannikvierling There is one regression on the TIP proofs: I had to replace one call to AIP by sorry in isaplanner/prop_49.  The problem is that Escargot now times out on that subproblem.
I think this happens because the hashcode of the expressions changed, and Escargot previously found the proof mostly by luck.

Can you fix that call to AIP?  If this is the only regression, I'd rather not spend much time fixing it.

jannikvierling:
@gebner I will remove this call to AIP and replace it by a manual proof.

PR #612, opened 2017-03-10
Clean up branch (Cut-intro)   Don't merge yet

Lettmann:


core/src/main/scala/at/logic/gapt/expr/typedLambdaCalculus.scala
```diff
@@ -105,6 +105,11 @@ abstract class Expr {
    */
   def find( exp: Expr ): List[HOLPosition] = getPositions( this, _ == exp )
 
+  /*
+  Tests whether an expression is a subexpression.
+  def contains( exp: Expr ): Boolean = {}
```
gebner: This seems like a reasonable addition.  You could implement it using the `find` method that's right above it.

Lettmann: Actually, that is the point why I mentioned it. I thought that there should be a better way of implementing this than using positions (given that positions should not be necessary to do so)

gebner:
@Lettmann Sorry, I know you didn't want this PR to be merged yet, but I need the cleaned-up version.

PR #618, opened 2017-03-15
Natural deduction

gebner:
@mschlaipfer @MarielleLouise I'm creating a PR so that I can assign it to the 2.6 milestone.


gebner:
@mschlaipfer @MarielleLouise If we want to merge this for 2.6, then you need to merge master.  There have been lots of changes to master in the meantime, and this branch no longer builds with them.

mschlaipfer:
I've pushed the test cases and removed the examples. I've asked Marielle to check that I didn't introduce a mistake in translating the test cases of the operators she's implemented. Afterwards this should be mergable.

mschlaipfer:
@gebner I've only seen your comment now. I've merged master before pushing yesterday. Commit  	#2e7dc71
Do I need to do something else?

core/src/main/scala/at/logic/gapt/proofs/nd/nd.scala
```diff
@@ -0,0 +1,1354 @@
...
+   * @return
+   */
+  def getLeftSequentConnector: SequentConnector = occConnectors( 0 )
+
+  /**
+   * The object connecting the lower and middle upper sequents.
+   *
+   * @return
+   */
+  def getmiddleSequentConnector: SequentConnector = occConnectors( 1 )
```
gebner: The "m" should be upper-case here.

core/src/main/scala/at/logic/gapt/proofs/nd/nd.scala
```diff
@@ -0,0 +1,1354 @@
...
+ * An NDProof that consist of the introduction of an equality.
+ * <pre>
+ *   ---------- eq:i
+ *    :- t = t
+ *
+ * </pre>
+ *
+ * @param t The term t.
+ */
+case class EqualityIntroRule( t: FOLTerm ) extends InitialSequent {
```
gebner: This should be `Expr` instead of `FOLTerm`.

core/src/main/scala/at/logic/gapt/proofs/nd/nd.scala
```diff
@@ -0,0 +1,1354 @@
...
+
+    val ( s, _ ) = eqFormula match {
+      case Eq( s, t ) => ( s, t )
+      case _          => throw NDRuleCreationException( s"Formula $eqFormula is not an equation." )
+    }
+
+    val repContext = replacementContext.abstractTerm( auxFormula )( s )
+
+    val formulaA = repContext.term.asInstanceOf[Formula]
+    val variablex = repContext.variable.asInstanceOf[FOLVar]
```
gebner: This should be `Var` instead of `FOLVar`.

gebner:
I quickly read over the code and added a few minor nitpicks.  @MarielleLouise Anything missing from your side?

One more thing that we should talk about is the location of the major premise in the elimination rules.  For the or-elimination rule it is the last premise, while for all other elimination rules it is the first premise.  Should we move the major premise of the or-elimination rule to the front as well?  What do you think?

shetzl:
> Should we move the major premise of the or-elimination rule to the front as well?

I am in favour of moving it to the front. It is the way this inference rule is usually presented.

tests/src/test/scala/at/logic/gapt/proofs/nd/NDTest.scala
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  "EqualityElim 4" in {
+    val c1 = LogicalAxiom( fof"!x0!x1 P(x2)" )
+    val c2 = LogicalAxiom( fof"x2=x1" )
+    val c3 = EqualityElimRule( c2, c1 )
+    c3.conclusion mustEqual Seq( fof"x2 = x1", fof"!x0 !x1 P x2" ) ++: Sequent() :+ fof"!x0 !x1_0 P x1"
+  }
+
+  "EqualityElim 4" in {
```
mschlaipfer: Should be EqualityElim 5

MarielleLouise:
@gebner I would just like to look at the tests and maybe add some more tests this evening, that is all. 

I also agree to move it to front so will do that too then.

PR #630, opened 2017-04-28
added extended test cases for induction elimination

jannikvierling:


gebner:
> Depending on how the tests run I will add more tests for different sizes of instance terms later on.

I agree, this seems like a good plan.

PR #633, opened 2017-05-24
LK to ND translation

mschlaipfer:


gebner:
As discussed on gitter, I have merged this PR for testing in the regression tests.  I will comment on the code next week.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -0,0 +1,566 @@
...
+        val i = tr.endSequent.indexOfPol( rightSubProof.endSequent( aux2 ), Polarity.InAntecedent )
+
+        val partialProof = nd.ProofBuilder.
+          c( tr ).
+          u( ImpIntroRule( _, i ) ).
+          c( tl ).
+          b( ImpElimRule( _, _ ) ).
+          qed
+
+        if ( p.endSequent.succedent.nonEmpty ) {
```
gebner: Since every case ends with a potential exchange, it might be simpler to move the exchange outside of the match to the end of the translate method.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -0,0 +1,566 @@
...
+          AndIntroRule( tl, tr )
+        } else {
+          val focusMain = p.endSequent.indexOfPol( p.mainFormula, Polarity.InSuccedent )
+          exchange( translate( proof, focusMain ), p.endSequent( focus ) )
+        }
+
+      case p @ OrLeftRule( leftSubProof, aux1, rightSubProof, aux2 ) =>
+
+        val tl =
+          if ( p.endSequent.succedent.nonEmpty ) {
```
gebner: This code snippet occurs quite a few times, maybe we should make it a tad shorter.  If I see this correctly, it should do the following:
 1. Translate the subproof with the parent index of the focus.  There are two special cases: the succedent is empty, or the focus does not have a parent.
 2. Move the focused formula into the ND-succedent.

For 1), we can write this as
```scala
translate(leftSubProof, if (p.endSequent.succedent.nonEmpty) p.getLeftSequentConnector.parent(focus, heuristicFocus) else heuristicFocus)
```
This still has the explicit special case if the succedent is empty.  Conceptually, I don't particularly like using `Suc(0)` if there is no formula in the succedent, I think an `Option[SequentIndex]` would be clearer.

For 2), I think it would be nice if we could unconditionally invoke exchange on the produced proofs.
```scala
// def exchange(p: NDProof, f: Option[Formula]): NDProof // optionally moves f to the succedent

val tl = exchange(translate(leftSubProof, ...), p.endSequent.get(focus))
```

I think it is important for readability that there are as few execution paths as possible.  If I see `exchange(translate(leftSubProof, ...), p.endSequent.get(focus))`, then I know that no matter what, we translate the left sub-proof and then exchange some formulas.  It is also immediately clear that the produced proof has the focused formula in the succedent.  With a case split, I need to consider each case separately and prove all these properties for each case individually.

core/src/main/scala/at/logic/gapt/proofs/nd/ProofBuilder.scala
```diff
@@ -0,0 +1,80 @@
...
+   * @return
+   */
+  def qed = proofStack match {
+    case Seq()      => throw new Exception( "Proof stack is empty." )
+    case p +: Seq() => p
+    case _          => throw new Exception( "There is more than one proof on the stack." )
+  }
+}
+
+object ProofBuilder extends ProofBuilder[NDProof]( Seq() )
```
gebner: There's no need to copy-paste the whole ProofBuilder code, it is already parametrically polymorphic in the type of proofs it constructs.  You can just instantiate it for ND proofs:
```scala
object ProofBuilder extends lk.ProofBuilder[NDProof](Seq())
```

core/src/main/scala/at/logic/gapt/proofs/nd/nd.scala
```diff
@@ -1053,6 +1077,30 @@ case class TheoryAxiom( mainFormula: Formula ) extends InitialSequent {
...
+object TheoryAxiom extends ConvenienceConstructor( "TheoryAxiom" ) {
+
+  /**
+   * Convenience constructor for ax, taking a context.
+   * Applies the axiom rule followed by 0 or more weakenings.
+   * <pre>
+   *    ------ax
+   *     :- A
+   *    ---------wkn*
+   *     Γ :- A
```
gebner: Oh well, this is definitely not the right translation for theory axioms from LK.  But it suffices as a stop-gap solution.

A typical theory axiom we have in LK proofs is for example `x+1<y :- x<y`.  Note that `:- x<y` is not only not an axiom, it is also not true (in the intended model).

Ideally, we'd translate this theory axiom as `Axiom(∀x ∀y (x+1<y ⊃ x<y))`, followed by ∀- and ⊃-elimination.  But unfortunately the general version of the axiom is not stored in the LK ProofLink rule, we only store the concrete instance (e.g. `1+1<3 ⊃ 1<3`) there.  The general version is available from the Context, however.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -0,0 +1,566 @@
...
+    } else {
+      assert( lk.endSequent.size == nd.endSequent.size )
+      assert( lk.endSequent.succedent.contains( nd.endSequent( Suc( 0 ) ) ) )
+      assert( lk.endSequent( focus ) == nd.endSequent( Suc( 0 ) ) )
+    }
+    assert( lk.endSequent.antecedent.forall( nd.endSequent.antecedent.contains( _ ) ) )
+    assert( lk.endSequent.succedent.filter( _ != nd.endSequent( Suc( 0 ) ) ).forall( x => nd.endSequent.antecedent.contains( Neg( x ) ) ) )
+  }
+
+  private def exchange( subProof: NDProof, mainFormula: Formula ): NDProof = {
```
mschlaipfer: The difference is in the placement of the negations. I sent you a document via email that shows the rules. I am aware that this is not elegant, but only realized the need for additional exchange rules when doing the equality rules.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -0,0 +1,566 @@
...
+            c( nd.LogicalAxiom( hof"-$r" ) ).
+            u( NegElimRule( _, subProof ) ).
+            u( BottomElimRule( _, mainFormula ) ).
+            qed
+        }
+
+        val i = pr2.endSequent.indexOfPolOption( negMain, Polarity.InAntecedent )
+        ExcludedMiddleRule( ax1, Ant( 0 ), pr2, i.get )
+      } else {
+        // TODO this case seems to be only hit by NegRight, so it should be handled there.
```
mschlaipfer: Not sure about this, maybe you can elaborate? I also just realized that the translated NegRight rule corresponds to `exchange3`.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -0,0 +1,566 @@
...
+          val t = translate( subProof, aux1 )
+          val il = t.endSequent.indexOfPolOption( hof"-$l", Polarity.InAntecedent )
+          nd.ProofBuilder.
+            c( nd.LogicalAxiom( l ) ).
+            c( t ).
+            b( ExcludedMiddleRule( _, Ant( 0 ), _, il.get ) ).
+            qed
+        } else {
+          val focusMain = p.endSequent.indexOfPol( p.mainFormula, Polarity.InSuccedent )
+          exchange( translate( proof, focusMain ), p.endSequent( focus ) )
```
mschlaipfer: We do something, but I admit it is a bit misleading. We translate `proof` and not `subProof` in the `else` branch, only changing the focus. Unless I am missing something. There is an example in lk_to_nd_examples.scala: `contractRightWithWrongFocus`. I didn't add it into the test suite for some reason.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
mschlaipfer: I agree and will look into refactoring this. I assume you mean `p.getLeftSequentConnector.parentOption(focus).getOrElse(heuristicFocus)`?

core/src/main/scala/at/logic/gapt/proofs/nd/ProofBuilder.scala
mschlaipfer: I see, the fact that it was in the `lk` package made me not even check that it is parametrized. I'd need to add the tertiary rule to the class inside the `lk` package, however. Unless there is some way to extend `lk.ProofBuilder` in Scala despite the constructor being private.

core/src/main/scala/at/logic/gapt/proofs/nd/nd.scala
mschlaipfer: I assume I can access the context as in `LKToExpansionProof`?

gebner: Yes, I don't think there's any way around it.

mschlaipfer: Could you give a concrete code example for setting up the context and a proof that requires accessing the context for correct translation?

Does the following make sense?

```
  ctx += Context.Sort( "i" )
  ctx += hoc"'<': i>i>o"
  ctx += hoc"'+': i>i>i"
  ctx += hoc"'1': i"
  ctx += ("ax", hos"x + 1 < y :- x < y")
  val lk = ProofLink( foc"th", hos"1 + 1 < 3 :- 1 < 3" )
```

gebner: Almost.  The last lines should be
```scala
val lk = ProofLink(le"ax 1 3", hos"1 + 1 < 3 :- 1 < 3")
ctx.check(lk) // just a sanity check
```

If you print `ctx`, then you should see the declaration of the `ax` constant (showing that it takes two arguments).

PR #635, opened 2017-06-08
LK to ND

mschlaipfer:
A bit of refactoring and some fixes for regression test failures.


core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -148,34 +142,33 @@ object LKToND {
...
         }
       }
     }
   }
 
-  private def translate( proof: LKProof, focus: SequentIndex ): NDProof = {
+  private def translate( proof: LKProof, focus: Option[SequentIndex] ): NDProof = {
 
-    assert( focus.isSuc )
+    if ( focus != None )
```
gebner: Options behave like containers with at most one element.  They have map, foreach, filter, flatMap, forall, etc.  Here you could write `assert(focus.forall(_.isSuc))`

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -196,14 +189,14 @@ object LKToND {
 
       case p @ WeakeningRightRule( subProof, formula ) =>
 
-        if ( p.mainFormula == p.endSequent( focus ) ) {
+        if ( p.mainFormula == p.endSequent( focus.get ) ) {
           // Pick arbitrary focus
           val heuristicIndex = Suc( 0 )
-          exchange( WeakeningRule( translate( subProof, heuristicIndex ), hof"-$formula" ), p.mainFormula )
+          exchange( WeakeningRule( translate( subProof, Some( heuristicIndex ) ), hof"-$formula" ), p.mainFormula )
```
gebner: What happens to heuristicIndex with `WeakeningRightRule(BottomAxiom, hof"p")`?

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -196,14 +189,14 @@ object LKToND {
...
         } else {
           // simply weaken with negated formula on the left
-          val focusMain = p.endSequent.indexOfPol( p.endSequent( focus ), Polarity.InSuccedent )
-          WeakeningRule( translate( subProof, focusMain ), hof"-$formula" )
+          val focusMain = p.endSequent.indexOfPol( p.endSequent( focus.get ), Polarity.InSuccedent )
```
gebner: This works by sheer accident since weakening adds the formula on the right and the indices for the formulas in the premise and conclusion coincide (you actually want the index in the premise):
```
:- a,b,c
------------ w:r
:- a,b,c,d
```
You 1) might as well pass the focus index unchanged, or 2) "follow" the index upwards using the sequent connectors: `focus.map(p.getSequentConnector.parent)`

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -223,26 +216,14 @@ object LKToND {
             qed
         } else {
           val focusMain = p.endSequent.indexOfPol( p.mainFormula, Polarity.InSuccedent )
```
gebner: As with WeakeningRight, looking up the index in p instead of subProof only works by accident.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
```diff
@@ -254,7 +235,7 @@ object LKToND {
           qed
 
         if ( p.endSequent.succedent.nonEmpty ) {
```
gebner: This pattern occurs in almost every rule (that is, exchanging if focus is non-empty).  Maybe we should add an extra exchange method that takes an optional formula as argument?
```scala
def exchange(subProof: NDProof, mainFormula: Option[Formula]): NDProof =
  mainFormula.map(exchange(subProof, _)).getOrElse(subProof)
```
Then we can write this pattern as:
```scala
exchange(partialProof, focus.map(p.endSequent.apply))
```
Or maybe even just once for all rules at the end of the method.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
mschlaipfer: I see your point, but it works as `BottomAxiom`, doesn't use focus. Similar with `NegLeftRule`.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToND.scala
mschlaipfer: This sound good. I will rewrite and try to move `exchange` to the end.

PR #640, opened 2017-06-28
Induction elimination

jannikvierling:


core/src/main/scala/at/logic/gapt/proofs/lk/reductiveCutElimination.scala
```diff
@@ -506,10 +503,18 @@ object leftRankReduction {
         Some( ForallLeftRule( cutSub, aNew, f, term, quant ) )
 
       case l @ ForallRightRule( subProof, a, eigen, quant ) if left.mainIndices.head != aux1 =>
-        val aux1Sub = l.getSequentConnector.parent( aux1 )
-        val cutSub = CutRule( l.subProof, aux1Sub, right, aux2 )
-        val aNew = cutSub.getLeftSequentConnector.child( a )
-        Some( ForallRightRule( cutSub, aNew, eigen, quant ) )
+        val regularCut @ CutRule( _, _, _, _ ) = regularize( cut )
```
gebner: This is just overly cautious. :smile:

In most cases we don't need to do any renaming at all, and we never need to rename the other side of the cut.  Regularization is "just" linear in the size of the proof, but it's still slow.

One nice idiom used in the substitution code is to add a second `ForallRightRule` case that handles the situation where we need to rename, and then call the function again with the fixed argument:
```scala
case l @ ForallRightRule( subProof, a, eigen, quant ) if left.mainIndices.head != aux1
    && freeVariables(cut.endSequent).contains(eigen) =>

  val newEigen = rename(eigen, freeVariables(cut.endSequent))
  val renaming = Substitution(eigen -> newEigen)
  apply(CutRule(left, aux1, ForallRightRule(renaming(subProof), a, newEigen, quant), aux2))

case l @ ForallRightRule( subProof, a, eigen, quant ) if left.mainIndices.head != aux1 =>
  // rank-reduce without renaming
```

gebner: I can't put a comment there, but with this change we can also remove the `regularize` call in `ReductiveCutElimination.apply`.

core/src/main/scala/at/logic/gapt/proofs/lk/Utils.scala
```diff
@@ -71,8 +71,14 @@ object isRegular {
    * @return true iff proof is regular.
    */
   def apply( proof: LKProof ): Boolean = {
-    val eigenVars = for ( Eigenvariable( v ) <- proof.treeLike.postOrder ) yield v
-    eigenVars == eigenVars.distinct
+    val eigenvariables: Seq[Var] = proof.subProofs.toSeq flatMap {
```
jannikvierling: Yes I think we should remove the `Eigenvariable` matcher. We could also modify it so that it returns a set of eigenvariables, but I think this would be rather inconvenient to use with `ForallRightRule` and `ExistsLeftRule`.

gebner: The matcher is only used here.  I think we can safely remove it.

gebner:
Are you changing anything before the weekend?  Either way you should merge it today so that it gets tested.

jannikvierling:
I will do the changes during the weekend.

PR #644, opened 2017-07-03
Induction elimination

jannikvierling:


gebner:
Should we merge this for the 2.7 release on Wednesday?  I'm a bit anxious since we didn't test it.

shetzl:
I rather tend to being cautious and to only merge it after the release.

PR #649, opened 2017-10-04
Schproofupdate

Ermine516:


core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+  }
+  //Finds the proof links within the given struct
+  object StructDependences {
+    def apply(
+      topSym:        String,
+      cutConfig:     HOLSequent,
+      currentStruct: Set[( ( Expr, Set[Var] ), Struct[Nothing] )],
+      foundCases:    Set[( String, HOLSequent )] ): Set[( String, HOLSequent )] =
+      currentStruct.fold( Set[( String, HOLSequent )]() )( ( w, e ) => {
+        val temp: Set[Struct[Nothing]] = SchematicLeafs( e.asInstanceOf[( Expr, Struct[Nothing] )]._2 ).fold( Set[Struct[Nothing]]() )( ( g, pb ) => {
```
gebner: Get rid of the `asInstanceOf`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+        ( m1.keySet ++ m2.keySet ) map { i => i -> ( m1.get( i ).toList ::: m2.get( i ).toList ) } toMap
+      else if ( m1.keySet.isEmpty && m2.keySet.nonEmpty )
+        m2.keySet map { i => i -> m2.get( i ).toList } toMap
+
+      else if ( m1.keySet.nonEmpty && m2.keySet.isEmpty )
+        m1.keySet map { i => i -> m1.get( i ).toList } toMap
+      else Map[K, List[V]]()
+
+    def mergeSet[K, V]( m1: Map[K, Set[V]], m2: Map[K, Set[V]] ): Map[K, Set[V]] =
+      if ( m1.keySet.nonEmpty && m2.keySet.nonEmpty )
```
gebner:
```scala
Map() ++ (for (k <- m1.keySet ++ m2.keySet)
  yield m1.getOrElse(k, Set()) ++ m2.getOrElse(k, Set()))
```
and similarly for the other methods.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+        m1
+      else Map[K, Set[V]]()
+  }
+
+  //Constructs clause sets modulo the cut configurations and proofs
+  object CutConfigProofClauseSetMaps {
+    def apply(
+      cutConfig:     HOLSequent,
+      currentStruct: Set[( ( Expr, Set[Var] ), Struct[Nothing] )] ): Map[HOLSequent, Set[( ( Expr, Set[Var] ), Set[SetSequent[Atom]] )]] = currentStruct.map( x => {
+      val theClauseSet = CharacteristicClauseSet( x._2, cutConfig ).asInstanceOf[Set[SetSequent[Atom]]]
```
gebner: That's just too many `asInstanceOf` calls and type ascriptions.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+
+      Map[HOLSequent, Set[( ( Expr, Set[Var] ), Set[SetSequent[Atom]] )]]( ( cutConfig, clauseSetNameVarsMatch ) )
+    } ).fold( Map[HOLSequent, Set[( ( Expr, Set[Var] ), Set[SetSequent[Atom]] )]]() )( ( aHOLIndexedMap, instance ) => {
+      instance.keySet.fold( Map[HOLSequent, Set[( Expr, Set[SetSequent[Atom]] )]]() )( ( variationMap, sequentInstances ) => {
+        if ( aHOLIndexedMap.keySet.contains( sequentInstances.asInstanceOf[HOLSequent] ) ) {
+          val thevalinw = instance.get( sequentInstances.asInstanceOf[HOLSequent] ) match {
+            case Some( x ) => x
+            case None      => Set[( ( Expr, Set[Var] ), Set[SetSequent[Atom]] )]()
+          }
+          val placeVariationsInMap = aHOLIndexedMap.get( sequentInstances.asInstanceOf[HOLSequent] ) match {
```
gebner: `.getOrElse(Set())`

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+    } )
+  }
+
+  def nat( i: Int, thevar: Var )( implicit ctx: Context ): Expr = {
+    val suc = ctx.get[Context.Constants].constants.getOrElse( "s", Const( "0", Ti ) )
+    if ( i > 0 ) Apps( suc, Seq( nat( i - 1, thevar ) ) )
+    else thevar
+  }
+
+  object InstantiateClauseSetSchema {
```
gebner: This should not be in the `SchematicClauseSet` object, probably best in another file.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+        case Apps( Const( "CL", _ ), _ ) => ( seqNotCL.asInstanceOf[Set[V]],
+          seqInCL.asInstanceOf[Set[V]] ++ Set[V]( y.asInstanceOf[V] ) )
+        case _ => ( seqNotCL.asInstanceOf[Set[V]] ++ Set[V]( y.asInstanceOf[V] ),
+          seqInCL.asInstanceOf[Set[V]] )
+      }
+    } ).asInstanceOf[( Set[V], Set[V] )]
+  }
+  //This object is specifically designed to read clause set terms
+  //which are constructed from proofs during struct construction
+  object ClauseTermReader {
```
gebner: @shetzl has strong opinions about this representation.  We should discuss this next week when he's back in the office.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+      val newclauses = Cout._1.map( x =>
+        if ( Cout._2.nonEmpty ) Cout._2.map( y => sequentCompose( x, y ) )
+        else Cout._1 )
+      if ( newclauses.nonEmpty )
+        newclauses.tail.fold( newclauses.head )( ( x, y ) => x ++ y )
+      else Set()
+    }
+    //This is a compose function made specifically for composing schematic clauses during clause set construction
+    def sequentCompose( S1: Sequent[Atom], S2: Sequent[Atom] ): Sequent[Atom] =
+      Sequent[Atom]( S1.antecedent.distinct ++ S2.antecedent.distinct, S1.succedent.distinct ++ S2.succedent.distinct )
```
gebner: `(S1 ++ S2).distinct`

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+  object ComposeClauseSets {
+    def apply(
+      C1: Set[Sequent[Atom]],
+      C2: Set[Sequent[Atom]] ): Set[Sequent[Atom]] = {
+      val Cout: ( Set[Sequent[Atom]], Set[Sequent[Atom]] ) = if ( C1.size > C2.size ) ( C1, C2 ) else ( C2, C1 )
+      val newclauses = Cout._1.map( x =>
+        if ( Cout._2.nonEmpty ) Cout._2.map( y => sequentCompose( x, y ) )
+        else Cout._1 )
+      if ( newclauses.nonEmpty )
+        newclauses.tail.fold( newclauses.head )( ( x, y ) => x ++ y )
```
gebner: There's no need to reimplement `flatMap`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+        if ( totalcount > oldcount ) ( totalcount, current._1, clauses )
+        else theCorrect
+      } )
+  }
+  //Takes two clause sets and composes them clause by clause without duplication
+  object ComposeClauseSets {
+    def apply(
+      C1: Set[Sequent[Atom]],
+      C2: Set[Sequent[Atom]] ): Set[Sequent[Atom]] = {
+      val Cout: ( Set[Sequent[Atom]], Set[Sequent[Atom]] ) = if ( C1.size > C2.size ) ( C1, C2 ) else ( C2, C1 )
```
gebner: If I read this correctly, this function does the same as:
```scala
if (C1.isEmpty) C2
else if (C2.isEmpty) C1
else for (c1 <- C1; c2 <- C2) yield (c1 ++ c2).distinct
```
Why do we special-case empty clause sets here?

core/src/main/scala/at/logic/gapt/proofs/ceres/StructCreators.scala
```diff
@@ -45,15 +48,16 @@ object StructCreators extends Logger {
   def toFormula[Data]( s: Struct[Data] ): Formula =
     And( CharacteristicClauseSet( s ).toSeq map ( _.toDisjunction ) )
 
-  def extract[Data]( p: LKProof ): Struct[Data] =
-    extract[Data]( p, p.endSequent.map( _ => false ) )( x => true )
-  def extract[Data]( p: LKProof, predicate: Formula => Boolean ): Struct[Data] =
-    extract[Data]( p, p.endSequent.map( _ => false ) )( predicate )
+  def extract[Data]( p: LKProof, ctx: Context ): Struct[Data] =
```
gebner: `Context` should be implicit here.

core/src/main/scala/at/logic/gapt/proofs/ceres/StructCreators.scala
```diff
@@ -130,18 +141,24 @@ object StructCreators extends Logger {
...
+  def handleAxiom[Data](
+    so:        HOLSequent,
+    cut_occs:  Sequent[Boolean],
+    ctx:       Context,
+    proofLink: String           = "" ): Struct[Data] = {
     //printf( "Axiom!" )
     //printf( cut_occs.toString )
 
     val cutanc_seq: HOLSequent = so.zipWithIndex.filter( x => cut_occs( x._2 ) ).map( _._1 )
+    /* println("this is important")
```
gebner: Please remove debugging code.

core/src/main/scala/at/logic/gapt/proofs/ceres/StructCreators.scala
```diff
@@ -130,18 +141,24 @@ object StructCreators extends Logger {
...
+          Plus[Data]( extract[Data]( upperProofLeft, new_occs1, ctx ), extract[Data]( upperProofRight, new_occs2, ctx ) )
         else
-          Times[Data]( extract[Data]( upperProofLeft, new_occs1 ), extract[Data]( upperProofRight, new_occs2 ), List() )
+          Times[Data]( extract[Data]( upperProofLeft, new_occs1, ctx ), extract[Data]( upperProofRight, new_occs2, ctx ), List() )
       case _ => throw new Exception( "Missing rule in StructCreators.extract: " + p.name )
     }
   }
 
-  def handleAxiom[Data]( so: HOLSequent, cut_occs: Sequent[Boolean] ): Struct[Data] = {
+  def handleAxiom[Data](
```
gebner: I think we should just make a separate `handleProofLink` method.  Dispatching on whether `proofLink` is the empty string or not is an ugly hack.

core/src/main/scala/at/logic/gapt/proofs/ceres/StructCreators.scala
```diff
@@ -155,8 +172,45 @@ object StructCreators extends Logger {
         val cutAncInAntecedent = cutanc_seq.antecedent.map( x => Dual[Data]( A( x, Nil ) ) )
         val cutAncInSuccedent = cutanc_seq.succedent.map( x => A[Data]( x ) )
         val structs: Vector[Struct[Data]] = cutAncInAntecedent ++ cutAncInSuccedent
-        val r = Times[Data]( structs, List[Data]() )
-        r
+        if ( !proofLink.matches( "" ) ) {
```
gebner: You want `proofLink != ""` or `proofLink.nonEmpty` here.  The `matches` method treats `""` as a regular expression, which seems slightly over-the-top here.

core/src/main/scala/at/logic/gapt/proofs/ceres/StructCreators.scala
```diff
@@ -155,8 +172,45 @@ object StructCreators extends Logger {
...
+          //This code matches positiions for terms passed through the proof links
+          val ( Apps( _, vs ), hs: HOLSequent ) = ctx.get[ProofNames].names.get( proofLink ) match {
+            case Some( ( ex, hs ) ) => ( ex, hs )
+            case None               => ( Const( "", Ti ), HOLSequent() )
```
gebner: Just let the code fail in cases that should not happen.  There is no point in producing wrong results, an exception is always easier to debug.

core/src/main/scala/at/logic/gapt/proofs/ceres/Struct.scala
```diff
@@ -147,6 +149,25 @@ object A {
   def apply[Data]( fo: Formula ): Struct[Data] = A[Data]( fo, Nil )
 }
 
+case class CLS[Data]( proof: String, config: HOLSequent, fv: Seq[Expr], data: List[Data] ) extends Struct[Data] { // Clause Set Symbol Struct
```
gebner: What does `fv` mean here?

core/src/main/scala/at/logic/gapt/proofs/lk/IsKSimple.scala
```diff
@@ -0,0 +1,34 @@
...
+import at.logic.gapt.expr.{ ->, Const, Expr, TBase, Ty, Var }
+import at.logic.gapt.proofs.Context.InductiveType
+
+object IsKSimple {
+  def apply( proof: LKProof ): Boolean = {
+    val base: Ty = TBase( "nat", List() )
+    val successor: Const = Const( "s", ->( base, base ) )
+    val zero: Const = Const( "0", base )
+    val nat = InductiveType( base, zero, successor )
+    val inducProofs: List[LKProof] = proof.subProofs.foldRight( List[LKProof]() )( ( a, z ) => a match {
```
gebner:
```scala
proofs.subProofs.collect {
  case p: InductionRule if p.indTy == nat.ty => p
}
```

core/src/main/scala/at/logic/gapt/proofs/lk/IsKSimple.scala
```diff
@@ -0,0 +1,34 @@
...
+      case p: InductionRule =>
+        if ( p.indTy == nat.ty ) p :: z
+        else z
+      case _ => z
+    } )
+    val succ: List[( List[Var], Expr )] = inducProofs.map { a =>
```
gebner:
```scala
for (p <- inducProofs; c <- p.cases
    if c.constructor.name == "s")
  yield (c.eigenVars, p.term)
```

core/src/main/scala/at/logic/gapt/proofs/lk/IsKSimple.scala
```diff
@@ -0,0 +1,34 @@
...
+        val InductionRule( c, _, t ) = a
+        c.foldRight( ( List[Var](), t ) )( ( b, z ) => {
+          b match {
+            case InductionCase( _, Const( "s", _ ), _, e, _ ) => ( e.toList, t )
+            case _ => z
+          }
+        } )
+      }
+    }
+    if ( succ.nonEmpty ) succ.foldRight( true )( ( a, z ) => a._1.contains( a._2 ) && a._2 == succ.head._2 && z )
```
gebner: `succ.forall { case (evs, t) => ... }`

core/src/main/scala/at/logic/gapt/proofs/lk/Utils.scala
```diff
@@ -27,6 +27,15 @@ object containsDefinitionRules {
     }
 }
 
+object EigenVariablesLK {
+  def apply( p: LKProof ): Set[Var] = p match {
```
gebner: You may want to include the eigenvariables from induction rules here as well.

examples/lk_to_nd_examples.scala
```diff
@@ -467,6 +467,7 @@ object negLeft extends Script {
   val nd = LKToND( lk, Some( focus ) )
 
   println( lk )
+
```
gebner: Unrelated change, please revert.

tests/src/test/scala/at/logic/gapt/proofs/lk/SchemaTest.scala
```diff
@@ -20,85 +24,314 @@ class SchemaTest extends Specification {
...
+      SCS.keySet.size must beEqualTo( 3 )
+    }
+    " Extracting the Schematic Characteristic Clause Set Checking number configurations" in {
+      val SCS = SchematicClauseSet( "omega", ctx ) match {
+        case Some( x ) => x
+        case None      => Map[String, Map[HOLSequent, Set[( Expr, Set[SetSequent[Atom]] )]]]()
+      }
+
+      SCS.keySet.fold( 0 )( ( vale, x ) => {
+        SCS.get( x.asInstanceOf[String] ) match {
```
gebner: Get rid of the `asInstanceOf`, it's hiding the completely TYPE-UNSAFE usage of `fold` here.  `foldLeft` has the type you're looking for.  Even better, DON'T USE FOLD AT ALL:
```scala
SCS.values.map(_.size).sum
```

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -316,13 +317,23 @@ object Context {
...
     def +( name: String, referencedExpression: Expr, referencedProof: LKProof ) =
       copy( components + ( ( name, components.getOrElse( name, Set() ) + ( referencedExpression -> referencedProof ) ) ) )
-
-    def find( name: Expr ): Iterable[( LKProof, Substitution )] =
-      for {
-        defs <- components.values
-        ( declName, defPrf ) <- defs
-        subst <- syntacticMatching( declName, name )
-      } yield ( defPrf, subst )
+    //error here
```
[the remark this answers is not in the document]
Ermine516: Nope, should take it out

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,532 @@
...
+        }
+      //Once we find the definition of the proof we construct the
+      //Structs for the given proof modulo the provided cut
+      //configuration
+      val currentProofsStructs: Set[( ( Expr, Set[Var] ), Struct[Nothing] )] =
+        CurrentProofsCases.map( x => {
+          val ( placeHolder: Expr, assocProof: LKProof ) = x
+
+          val ancestorPositions = FindAncestors( assocProof.endSequent, cutConfig )
+          ( ( placeHolder, EigenVariablesLK( assocProof ) ), StructCreators.extract( assocProof, ancestorPositions, ctx )( _ => true ) )
```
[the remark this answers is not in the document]
Ermine516: This is a really complex problem, essentially clause set schema need a special regularization. I will explain more tomorrow.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -316,13 +317,23 @@ object Context {
...
+    def find( name: Expr ): Option[( LKProof, Substitution )] = {
```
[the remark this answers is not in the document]
Ermine516: The idea here concerns the Ndiffschema I added which has multiple basecases.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
Ermine516: :) I hope we could come up with a better one, I really dislike it as well.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
Ermine516: I don't remember why, but when It is removed some test fail.

core/src/main/scala/at/logic/gapt/proofs/ceres/Struct.scala
Ermine516: freevariables

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+      // For each CLS term we compute the Struct and merge the results
+      Some( clauseSetDependencies.map( x => SchematicStruct( x._1, x._2, foundCases ++ clauseSetDependencies - x +
+        ( topSym -> theActualConfig ) ).getOrElse( { throw new Exception( "Struct could not be built " ) } ) ).foldLeft( Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )]() )( ( x, y ) => x ++ y ) ++ currentProofStruct )
+    } else None
+  }
+}
+//Allows the construction of instances of schematic structs
+object InstanceOfSchematicStruct {
+  def apply[Data](
+    topSym:    Struct[Nothing],
```
gebner: This should be `topSym: CLS`, right?

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+// a Struct  and the expression of the case of the inductive definition.
+object SchematicStruct {
+  def apply( topSym: String, cutConfig: Sequent[Boolean] = Sequent[Boolean](),
+             foundCases: Set[( String, Sequent[Boolean] )] = Set[( String, Sequent[Boolean] )]() )( implicit ctx: Context ): Option[Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )]] = {
+    //If the context does not contain topSym then we cannot construct the struct and return None.
+    if ( ctx.get[ProofDefinitions].components.keySet.contains( topSym ) ) {
+      //If no cut config is provided we construct the default config
+      val theActualConfig = if ( cutConfig.isEmpty ) {
+        val ( _, theSeq ) = ctx.get[ProofNames].names.getOrElse( topSym, { throw new Exception( "Unhandled case: " + topSym ) } )
+        Sequent[Boolean]( theSeq.antecedent.map { _ => false }, theSeq.succedent.map { _ => false } )
```
gebner:
```scala
theSeq.map(_ => false)
```

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+      // We construct the struct for the given proof modulo the cutConfig
+      val currentProofStruct: Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )] =
+        ctx.get[ProofDefinitions].components.getOrElse( topSym, Set() ).map {
+          case ProofDefinition( placeHolder: Expr, _, assocProof: LKProof ) =>
+            ( CLS( placeHolder, theActualConfig ),
+              //need to write a predicate for cut ancestry
+              ( StructCreators.extract( assocProof, theActualConfig )( { _ => true }, ctx ), EigenVariablesLK( assocProof ) ) )
+        }.toMap
+      //After constructing the struct we need to find all CLS terms
+      val clauseSetDependencies = currentProofStruct.foldLeft( Set[( String, Sequent[Boolean] )]() )( ( w, e ) => {
```
gebner: Use `flatMap` instead of `foldLeft`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+        w ++ SchematicLeafs( e._2._1 ).foldLeft( Set[Struct[Nothing]]() )( ( g, pb ) => {
```
gebner: Use `filter` instead of `foldLeft`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+          val CLS( Apps( Const( pf, _ ), _ ), ccon ) = pb
+          if ( foundCases.contains( ( pf, ccon ) ) ) g else g + pb
+        } ).foldLeft( Set[( String, Sequent[Boolean] )]() )( ( y, a ) => {
+          val CLS( Apps( Const( pf, _ ), _ ), ccon ) = a
+          if ( pf.matches( topSym ) && ccon.equals( theActualConfig ) ) Set( ( pf, ccon ) ) else y ++ Set( ( pf, ccon ) )
```
gebner: This line discards a random subset of what we have constructed so far, is this right?

Otherwise, just use `map` instead of `foldLeft`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+        } )
+      } )
+      // For each CLS term we compute the Struct and merge the results
+      Some( clauseSetDependencies.map( x => SchematicStruct( x._1, x._2, foundCases ++ clauseSetDependencies - x +
+        ( topSym -> theActualConfig ) ).getOrElse( { throw new Exception( "Struct could not be built " ) } ) ).foldLeft( Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )]() )( ( x, y ) => x ++ y ) ++ currentProofStruct )
```
gebner: Use `flatten` instead of `foldLeft`.  Or even better, use `.flatMap` instead of `.map.foldLeft`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+object InstanceOfSchematicStruct {
+  def apply[Data](
+    topSym:    Struct[Nothing],
+    sss:       Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )],
+    sigma:     Substitution,
+    usedNames: Set[Var]                                            = Set[Var]() )( implicit ctx: Context ): Struct[Nothing] = {
+    val starterStruct = sss.getOrElse( topSym, { throw new Exception( "Not in Struct: " + topSym ) } )
+    val CLS( Apps( _, lex ), _ ) = topSym
+    // check if the Struct is empty or does not match sigma
+    if ( !sigma.domain.equals( freeVariables( lex ) ) ) EmptyPlusJunction()
```
gebner: So if I pass the substitution `x -> s(0)` I get an instance of the struct, but if I pass `x -> s(0), y -> s(0)` then everything is discarded?  This seems counterintuitive at best, what is the rationale for this behavior?

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+        ( reClause, nameVar ) => Set[Var]( Var( reClause._1._1.fresh( nameVar.name ), nameVar.ty ) ).map( newVar =>
+          ( ( reClause._1._1, reClause._1._2 + newVar ), RegularizeStruct( reClause._2, nameVar, newVar ) ) ).head ) ).map( x => ( x._1._2, x._2 ) ).head
+      //we now unfold the dependencies
+      InstantiateStruct( regularizedStruct._2, sigma, if ( freeVariables( lex.head ).size == 1 ) freeVariables( lex.head ).head
+      else Var( "", TBase( "nat" ) ), sss, usedNames ++ regularizedStruct._1 )
+    }
+  }
+}
+object RegularizeStruct extends StructVisitor[Struct[Nothing], List[Var]] {
+  def apply( theStruct: Struct[Nothing], nameVar: Var, newVar: Var ): Struct[Nothing] = {
```
gebner: I think this function should not be called "regularize".  It just renames one variable by another.  And for efficiency's sake, I'd use a `Map[Var, Var]` here so that we can rename multiple variables at once.

(Even better, write a `ClosedUnderReplacement[Struct]` instance.  Then you can use `TermReplacement(theStruct, nameVar, newVar)` or `TermReplacement(theStruct, theMap)`.)

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+    val Transform = StructTransformer[Struct[Nothing], List[Var]](
+      aF, { ( x, y, _ ) => Plus[Nothing]( x, y ) }, EmptyPlusJunction(), { ( x, y, _ ) => Times[Nothing]( x, y ) },
+      EmptyTimesJunction(), { ( x, _ ) => Dual[Nothing]( x ) }, cF )
+    recurse( theStruct, Transform, List[Var]( nameVar, newVar ) )
+  }
+  def aF[Data]( f: Formula, vars: List[Var] ): Struct[Data] = A( f.find( vars.head ).foldLeft( f )( ( ff, pos ) => ff.replace( pos, vars.tail.head ) ), List() )
+  def cF[Data]( pn: Expr, cc: Sequent[Boolean], vars: List[Var] ): Struct[Data] = {
+    val Apps( name, vs ) = pn
```
gebner:
```scala
val List(nameVar, newVar) = vars   // why on earth is this even a list???
CLS(TermReplacement(pn, nameVar, newVar), cc)
```

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+    val newVs = vs.map( f => sigma( sigma.domain.foldLeft( f )( ( subform, varsig ) =>
+      if ( varsig.ty.equals( TBase( "nat" ) ) )
+        subform.find( natMaker( 1, varsig ) ).foldLeft( subform )( ( nrepl, curpos ) => ctx.get[ProofNames].names.get( name ) match {
+        case Some( proofName ) =>
+          val Apps( _, lsymPN ) = proofName._1
+          if ( param.equals( lsymPN.head ) ) nrepl else if ( varsig.equals( param ) ) nrepl.replace( curpos, varsig ) else nrepl
+        case None => nrepl
+      } )
+      else subform ) ) )
+    val CLS( Apps( n, vs2 ), ccc ) = sss.keySet.foldLeft( EmptyPlusJunction().asInstanceOf[Struct[Nothing]] )( ( ret, theV ) => {
```
gebner: Use `find` + `getOrElse` instead of `foldLeft`.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+        ( topSym -> theActualConfig ) ).getOrElse( { throw new Exception( "Struct could not be built " ) } ) ).foldLeft( Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )]() )( ( x, y ) => x ++ y ) ++ currentProofStruct )
+    } else None
+  }
+}
+//Allows the construction of instances of schematic structs
+object InstanceOfSchematicStruct {
+  def apply[Data](
+    topSym:    Struct[Nothing],
+    sss:       Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )],
+    sigma:     Substitution,
```
gebner: After reading a bit through the code, I think this function should not take a substitution as argument since it can be computed from `topSym`.  Furthermore, I don't think we should require that `topSym.proof` is literally in the domain of `sss`, this should simplify `InstantiateStruct` significantly.

```scala
val (starterStruct, sigma) =
  (for {
    (sssCls, sssStruct) <- sss
    sigma <- syntacticMatching(sssCls.proof, topSym.proof)
  } yield (sssStruct, sigma)).head
```
Does this look reasonable?

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+  def aF[Data]( f: Formula, info: ( Substitution, Var, Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )], Set[Var] ) )( implicit ctx: Context ): Struct[Data] = {
+    val ( sigma, _, _, _ ) = info
+    A( sigma( sigma.domain.foldLeft( f )( ( subForm, varSig ) =>
+      ( if ( varSig.ty.equals( TBase( "nat" ) ) ) subForm.find( natMaker( 1, varSig ) )
+      else subForm.find( varSig ) ).foldLeft( subForm )( ( nRepl, curPos ) => nRepl.replace( curPos, varSig ) ) ) ), List() )
+  }
+  def cF[Data]( pn: Expr, cc: Sequent[Boolean], info: ( Substitution, Var, Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )], Set[Var] ) )( implicit ctx: Context ): Struct[Nothing] = {
+    val ( sigma, param, sss, usedNames ) = info
+    val Apps( Const( name, _ ), vs ) = pn
```
gebner: See my comment above at `InstanceOfSchematicStruct`.  I believe this function here could just be `InstanceOfSchematicStruct(CLS(sigma(pn), ccc), sss, usedNames)`.

core/src/main/scala/at/logic/gapt/expr/fol/utils.scala
```diff
@@ -233,3 +234,28 @@ object naive extends CountingFormulas {
   }
 
 }
+object mergeMapToListValues {
```
gebner: These two functions are unused, please remove them.  Besides, they shouldn't be in the `fol` package anyhow.

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  def cF( pn: Expr, cc: Sequent[Boolean], mn: Map[( ( String, Option[Expr] ), Sequent[Boolean] ), String] ): Formula = {
+    val Apps( Const( name, _ ), vs ) = pn
+    val thefirst = vs.headOption.getOrElse( { throw new Exception( "Should not be empty" ) } )
+    val result: Option[Expr] = freeVariables( thefirst ).headOption
+    Atom( mn.getOrElse( ( ( name, result ), cc ), { throw new Exception( "Should be in map" ) } ) + "PR", vs )
+  }
+  //assuming NNFCNF
+  private def QuantIntroForAll( f: Formula, qType: QuantifierC, evar: Set[Var] ): Formula = f match {
```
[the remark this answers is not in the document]
Ermine516: yes... at one point it wasn't :)

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,191 @@
...
+      val ( one, ( two, three ) ) = y
+      val theset = x.getOrElse( one, Set() ) ++ Set( ( two, three ) )
+      x ++ Map( ( one, theset ) )
+    } ) )
+  }
+  private def structNames( sss: Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )] ): Map[( ( String, Option[Expr] ), Sequent[Boolean] ), String] =
+    sss.keySet.map {
+      case CLS( Apps( Const( name, _ ), vs ), cc ) =>
+        val thefirst = vs.headOption.getOrElse( { throw new Exception( "Should not be empty" ) } )
+        val result: Option[Expr] = freeVariables( thefirst ).headOption
```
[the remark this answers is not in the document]
Ermine516: I added a convention that the first argument is the free parameter over which the current recursion is defined. I tried to do without this convention, but it made things too difficult.

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,125 @@
...
+import at.logic.gapt.expr.fol.natMaker
+import at.logic.gapt.proofs.Context.{ ProofDefinition, ProofDefinitions, ProofNames }
+import at.logic.gapt.proofs.lk.{ EigenVariablesLK, LKProof }
+import at.logic.gapt.proofs.{ Context, Sequent }
+
+//Idea behind the type is for each proof symbol we have a  Map,  which maps configurations to a
+// a Struct  and the expression of the case of the inductive definition.
+object SchematicStruct {
+  def apply( topSym: String, cutConfig: Sequent[Boolean] = Sequent[Boolean](),
+             foundCases: Set[( String, Sequent[Boolean] )] = Set[( String, Sequent[Boolean] )]() )( implicit ctx: Context ): Option[Map[Struct[Nothing], ( Struct[Nothing], Set[Var] )]] = {
```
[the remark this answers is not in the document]
Ermine516: I get an error when I type in CLS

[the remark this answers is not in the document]
gebner: Ah yes, at the moment this is still `Map[CLS[Nothing], (Struct[Nothing], Set[Var])]`.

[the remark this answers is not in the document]
Ermine516: Fixed it now. Changed everywhere

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
Ermine516: I need to look at it in more detail.

core/src/main/scala/at/logic/gapt/proofs/Context.scala
```diff
@@ -515,6 +567,17 @@ object Context {
       }
       PrimRecFun( c, nArgs, recIdx, eqns.toVector )
     }
+    def apply( PRF: Set[( Const, Seq[String] )] )( implicit ctx: Context ): primRecFunBatch = {
```
gebner: Sorry, I was probably unclear and caused a slight misunderstanding: the class names (like PrimRecFunBatch) were fine, I was taking about function parameters (like PRF here).  This is the typical style in GAPT:
```scala
def apply(prf: Set[(Const, Seq[String])])(implicit ctx: Context): PrimRecFunBatch = ...
```

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
Ermine516: :D This if statement actually does not do anything I removed it and replaced it with map

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
Ermine516: I think we should talk about this today, I am not familiar with these parts of the code.

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,166 @@
...
+        val cutConfigChars = cc.map( b => if ( b ) 'T' else 'F' )
+        val newName: String = name + "S" ++ cutConfigChars.succedent + "A" ++ cutConfigChars.antecedent
+        ( ( ( name, result ), cc ), newName )
+    }.toMap
+  private def addToContextAsPRDefs( ChF: Map[Expr, Set[( Expr, Expr )]] )( implicit ctx: MutableContext ): Unit = {
+    val prDefsForContext = for ( ( pred, _ ) <- ChF ) yield ( pred.asInstanceOf[Const], Seq( ChF( pred ).foldLeft( ( Some[Expr]( Atom( "", List() ) ), Some[Expr]( Atom( "", List() ) ) ) ) {
```
gebner: Is there a reason this function can't just be:
```scala
ctx += PrimRecFun {
  ChF.map { case (pred, eqns) =>
    eqns.map { case (lhs, rhs) => lhs === rhs }
  }
}
```

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,166 @@
...
+        throw new Exception( "?????" )
+      } )
+      val Atom( Const( name, _ ), vs ) = x
+      val newEx = Const( name, FunctionType( To, vs.map { _.ty } ) ).asInstanceOf[Expr]
+      ( ( name.substring( 0, name.length - 2 ), newEx ), ( newEx, ( Apps( newEx, vs: _* ), one ) ) )
+    } ).toList.unzip
+    val namesdis = namecha.toSet
+    addToContextAsPRDefs( nextRes.map {
+      case ( x, ( y, z ) ) =>
+        ( x, ( y, namesdis.foldLeft( z: Expr )( ( w, r ) => {
```
gebner:
```scala
val namedis = namecha.toMap
(x, (y, TermReplacement(z, {case Const(n, _) if namedis.contains(n) => namedis(n)})))
```

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,166 @@
...
+          TermReplacement( w, { case Const( n, _ ) if n == r._1 => r._2 } )
+        } ) ) )
+    }.foldLeft( Map[Expr, Set[( Expr, Expr )]]() ) {
```
gebner:
```scala
.groupBy(_._1)
.map { case (pred, eqns) => (pred, eqns.map(_._2).toSet) }
```

core/src/main/scala/at/logic/gapt/proofs/ceres/SchematicClauseSet.scala
```diff
@@ -0,0 +1,60 @@
...
+      Some( clauseSetDependencies.map( x => SchematicStruct( x._1, x._2, foundCases ++ clauseSetDependencies - x +
+        ( topSym -> theActualConfig ) ).getOrElse( { throw new Exception( "Struct could not be built " ) } ) ).foldLeft( Map[CLS, ( Struct, Set[Var] )]() )( ( x, y ) => x ++ y ) ++ currentProofStruct )
+    } else None
+  }
+}
+//Allows the construction of instances of schematic structs
+object InstanceOfSchematicStruct {
+  def apply( topSym: CLS, sss: Map[CLS, ( Struct, Set[Var] )], usedNames: Set[Var] = Set[Var]() )( implicit ctx: Context ): Struct = {
+    val ( starterStruct, sigma ) = ( for {
+      ( sssCls, sssStruct ) <- sss
```
gebner: I completely forgot about this, but I think we need to add an `if sssCls.config == topSym.config` here so that we pick the struct for the right cut-configuration.

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,148 @@
...
+
+private object Support {
+  def apply(
+    SCS: Map[CLS, ( Struct, Set[Var] )],
+    stT: StructTransformer[Formula, Map[( String, Sequent[Boolean] ), String]] ): Map[Formula, ( Formula, Set[Var] )] = {
+    val names = structNames( SCS )
+    SCS.keySet.map( x => {
+      val CLS( Apps( Const( name, _ ), vs ), cc ) = x
+      val ( one, two ) = SCS( x )
+      ( Atom( names( ( name, cc ) ) + "PR", vs ), ( constructingForm( one, names, stT ), two ) )
```
gebner: What is the rationale for adding a PR suffix?  We are already adding the `SFATF` suffix, which should be enough to disambiguate it from the proof name.

If we can drop the PR suffix, then we don't need to do any renaming in the `add` method below.

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,119 @@
...
+    case And( x, y )                             => new QuantifierHelper( ExistsC ).Block( evar.intersect( freeVariables( And( x, y ) ) ).toSeq, And( x, y ) )
+    case Atom( _, _ )                            => new QuantifierHelper( ExistsC ).Block( evar.intersect( freeVariables( f ) ).toSeq, f )
+    case Top()                                   => Top()
+    case Bottom()                                => Bottom()
+    case Neg( Neg( x ) )                         => QuantIntroExists( x, evar )
+    case Neg( x )                                => Neg( QuantIntroExists( x, evar ) )
+  }
+  def add( chF: Map[Formula, ( Formula, Set[Var] )], qType: QuantifierC )( implicit ctx: MutableContext ): Unit = {
+    val preRes = for ( ( x, ( form, vars ) ) <- chF ) yield ( x, if ( qType.equals( ForallC ) ) QuantIntroForAll( form, vars ) else QuantIntroExists( form, vars ) )
+    val changeRes = for {
```
gebner: Now that the PR suffix is gone, we can remove the renaming here.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+          if ( proof.endSequent.equals( y._2 ) ) y else x )
+        withIdentitySequentConnector( ProofLink( result._1, proof.endSequent ) )
+      case _ => super.recurse( proof, arg )
+    }
+  }
+}
+object ArithmeticInductionToSchema {
+  def apply( proof: LKProof, pe: Expr )( implicit ctx: MutableContext ): Unit = {
+    val Apps( Const( nameT, tpp ), _ ) = pe
+    val newNames: NameGenerator = rename.awayFrom(
```
gebner:
```scala
val newNames = ctx.newNameGenerator
newNames.fresh(nameT) // mark nameT as used
```
You could also declare the proof name `pe` before.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+      ctx.get[ProofNames].names.keySet.map( x => {
+        val Apps( Const( na, t ), _ ) = ctx.get[ProofNames].names.getOrElse( x, {
+          throw new Exception( "not defined" )
+        } )._1
+        Const( na, t )
+      } ) + Const( nameT, tpp ) )
+    val result = proof.subProofs.foldLeft( List[( Expr, ( Abs, LKProof ) )]() )( ( x, y ) => y match {
```
gebner: Use `collect` instead of `foldLeft`.

Please use a more descriptive name than `result`.  How about `subProofNames`?

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+      case InductionRule( _, form, typeTerm ) =>
+        val es = y.endSequent
+        val fv = typeTerm :: freeVariables( form ).toList
+        val ptype = FunctionType( typeTerm.ty, fv.map( _.ty ) )
+        val name = Const( newNames.fresh( "P" ), ptype )
+        val proofname = Apps( name, fv )
+        ctx += name
+        ctx += Context.ProofNameDeclaration( proofname, es )
+        ( proofname, ( form, y ) ) :: x
```
gebner: BTW, Scala has triples as well: `(proofname, form, y)`.  (also quadruples, quintuples, ..., 22-tuples)

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+      case _ => x
+    } )
+    val resProof: LKProof = CreateASchemaVersion( proof, ( result.unzip._1, ctx ) )
+    if ( ctx.get[ProofNames].lookup( pe ).isEmpty ) {
+      ctx += Const( nameT, tpp )
+      ctx += Context.ProofNameDeclaration( pe, resProof.endSequent )
+    }
+    ctx += Context.ProofDefinitionDeclaration( pe, resProof )
+    result.foreach( x => {
```
gebner: Use pattern-matching here:
```scala
result.foreach { case (proofName, (Abs(_,form), InductionRule(cases, _, term))) =>
  ...
}
```

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+      val Abs( _, form ) = x._2._1
+      val InductionRule( cases, _, term ) = x._2._2
+      cases.foreach( y => {
```
gebner: Again, use pattern-matching.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+        val InductionCase( p, _, hy, ev, con ) = y
+        val sigma = syntacticMatching( form, p.endSequent( con ) ).get
+        val endSequentLeft = ctx.get[ProofNames].lookup( x._1 ).getOrElse( { throw new Exception( "Proof not defined" ) } )
+        val finProof = if ( ev.nonEmpty ) {
+          //*********************************************************************************************
+          val newseq = Sequent(
+            endSequentLeft.antecedent.map( e =>
```
gebner: Use `endSequentLeft.map` instead of mapping separately over antecedent and succedent.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+              e.find( term ).foldLeft( e )( ( z, w ) => {
+                if ( freeVariables( e ).contains( term.asInstanceOf[Var] ) )
```
gebner: This won't work, in general `term` is not a variable (not even in the case of natural numbers).

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+      ctx.get[ProofNames].names.keySet.map( x => {
+        val Apps( Const( na, t ), _ ) = ctx.get[ProofNames].names.getOrElse( x, {
+          throw new Exception( "not defined" )
+        } )._1
+        Const( na, t )
+      } ) + Const( nameT, tpp ) )
+    val result = proof.subProofs.foldLeft( List[( Expr, ( Abs, LKProof ) )]() )( ( x, y ) => y match {
+      case InductionRule( _, form, typeTerm ) =>
+        val es = y.endSequent
+        val fv = typeTerm :: freeVariables( form ).toList
```
gebner: This will only work if `typeTerm` is a variable.  The induction rule in gapt can have an arbitrary term in its conclusion (e.g., you can prove `:- φ(2+3)` from `:- φ(0)` and `φ(x) :- φ(s(x))`).  So I think we need to generalize the sequent here when we declare a proof name.  In the example, we would pick a fresh variable `y` and declare a proof name for the generalized sequent `:- φ(y)` (instead of `:- φ(2+3)`).

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+    ctx += Context.ProofDefinitionDeclaration( pe, resProof )
+    result.foreach( x => {
+      val Abs( _, form ) = x._2._1
+      val InductionRule( cases, _, term ) = x._2._2
+      cases.foreach( y => {
+        val InductionCase( p, _, hy, ev, con ) = y
+        val sigma = syntacticMatching( form, p.endSequent( con ) ).get
+        val endSequentLeft = ctx.get[ProofNames].lookup( x._1 ).getOrElse( { throw new Exception( "Proof not defined" ) } )
+        val finProof = if ( ev.nonEmpty ) {
```
gebner: I think the proper way to do this in general is to fold[1] over the induction hypotheses `hy`, and for each induction hypothesis add a cut with the appropriate proof link (you should be able to get the substitution via matching, just as you got `sigma`) followed by a contraction.

[1] This is one of the few cases where I am happy with a fold. 😄

core/src/main/scala/at/logic/gapt/proofs/ceres/CharacteristicFormula.scala
```diff
@@ -0,0 +1,112 @@
...
+    case Top()                                   => Top()
+    case Bottom()                                => Bottom()
+    case Neg( Neg( x ) )                         => QuantIntroExists( x, evar )
+    case Neg( Atom( _, _ ) )                     => new QuantifierHelper( ExistsC ).Block( evar.intersect( freeVariables( f ) ).toSeq, f )
+    case Neg( x )                                => Neg( QuantIntroExists( x, evar ) )
+  }
+  def add( chF: Map[Formula, ( Formula, Set[Var] )], qType: QuantifierC )( implicit ctx: MutableContext ): Unit =
+    addToContextAsPRDefs( {
+      for {
+        (
```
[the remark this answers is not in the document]
gebner: BTW, isn't the whole repacking of the `newEx` completely pointless.  Can't we just do `(f @ Atom(newEx, vs), (form, vars)) <- chF`?

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,54 @@
...
+import at.logic.gapt.proofs.Context.ProofNames
+import at.logic.gapt.proofs.{ Context, MutableContext, SequentConnector }
+
+object CreateASchemaVersion extends LKVisitor[MutableContext] {
+  override protected def recurse( p: LKProof, ctx: MutableContext ): ( LKProof, SequentConnector ) = {
+    val newNames = ctx.newNameGenerator
+    p match {
+      case proof@InductionRule( casesI, form, typeTerm  ) =>
+        val formNorm =  BetaReduction.betaNormalize(form( typeTerm ))
+        val newVarForDef = Var( newNames.fresh( "x" ), typeTerm.ty )
```
gebner: `newNames` generates names that are fresh for the context, i.e., they are distinct from the defined constants.  Here we want a variable name that is not free in the sequent, so it's probably better to do `rename(Var("x", typeTerm.ty), freeVariable(proof.conclusion))`.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,54 @@
...
+        val es = proof.endSequent.map( f => typeTerm match {
```
gebner: In general, `TermReplacement` is a good idea if you want to replace constants uniformly by something else.  Here, we want to replace *some* occurrences of a term by a variable.  Luckily, we already have `form` so we don't need to do any matching.  It should be enough to just do `proof.endSequent.update(proof.mainIndices.head, BetaReduction.betaNormalize(form(newVarForDef).asInstanceOf[Formula]))` here.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,54 @@
...
+        val fv = freeVariables( TermReplacement( formNorm, typeTerm, newVarForDef ) ).toList
+        val name = Const( newNames.fresh( "Proof" ), FunctionType( typeTerm.ty, fv.map( _.ty ) ) )
+        val proofName = Apps( name, fv )
+        ctx += name
+        ctx += Context.ProofNameDeclaration( proofName, es )
+        casesI.foreach {
+          case InductionCase( subproof, _, hy, _, con ) =>
+            val sigma = syntacticMatching(formNorm , subproof.endSequent( con ) ).get
+            val endSequentLeft = ctx.get[ProofNames].lookup( proofName ).getOrElse( { throw new Exception( "Proof not defined" ) } )
+            val finProof = if ( hy.nonEmpty ) hy.foldLeft( subproof )( ( outputProof, hypoth ) => {
```
gebner: The case distinction on `hy.nonEmpty` seems unnecessary here.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,54 @@
...
+              val outputSeq = sigma( endSequentLeft.replaceAt( con, subproof.endSequent( hypoth ) ) )
+              ContractionMacroRule( CutRule( ProofLink( syntacticMatching( formNorm, subproof.endSequent( hypoth ) ).get( proofName ), outputSeq ), outputSeq.indexOf( outputProof.endSequent( hypoth ) ), outputProof, hypoth ), sigma( endSequentLeft ) )
+            } )
+            else {
+              val newante = endSequentLeft.antecedent.filter( t => subproof.endSequent.indexOfOption( t ).isEmpty && ( !t.contains( typeTerm ) || !freeVariables( t ).contains( typeTerm.asInstanceOf[Var] ) ) )
+              val newsuc = endSequentLeft.succedent.filterNot( t => subproof.endSequent.indexOfOption( t ).isEmpty && ( !t.contains( typeTerm ) || !freeVariables( t ).contains( typeTerm.asInstanceOf[Var] ) ) )
+              WeakeningRightMacroRule( WeakeningLeftMacroRule( subproof, newante ), newsuc )
```
gebner: Why do we need to add weakenings here?

Ermine516: The point of weakening is to match the basecase with the sequent of the stepcase. When the hypothesis is empty that means we are in a basecase, right? This is why I made the case distinction.

gebner: I still don't see why this is necessary.  For the proof definition, we require that the end-sequent of the proof is a submultiset of the declared sequent.  If it works with weakenings, it should work even better without.

Ermine516: I committed the code without weakening. It is breaking for some reason.

gebner: Oops, the check in `ProofDefinitionDeclaration` was completely wrong--it checked whether the declared sequent was a submultiset of the conclusion of the proof definition, i.e., the other way around.  That's also why the weakening made it work.  I just pushed a fix.

gebner: Note: `SchemaTest` still throws an exception.  I just fixed the check in the context.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,50 @@
...
+import at.logic.gapt.expr._
+import at.logic.gapt.proofs.Context.ProofNames
+import at.logic.gapt.proofs.{ Context, MutableContext, SequentConnector }
+
+object CreateASchemaVersion extends LKVisitor[MutableContext] {
+  override protected def recurse( p: LKProof, ctx: MutableContext ): ( LKProof, SequentConnector ) = {
+    val newNames = ctx.newNameGenerator
+    p match {
+      case proof @ InductionRule( casesI, form, typeTerm ) =>
+        val formNorm = BetaReduction.betaNormalize( form( typeTerm ) )
```
gebner: I don't think we want to compute `formNorm`.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,50 @@
...
+        val newVarForDef = rename( Var( "x", typeTerm.ty ), freeVariables( proof.conclusion ) )
+        val es = proof.endSequent.updated( proof.mainIndices.head, BetaReduction.betaNormalize( form( newVarForDef ).asInstanceOf[Formula] ) )
+        val fv = freeVariables( TermReplacement( formNorm, typeTerm, newVarForDef ) ).toList
```
gebner: Get rid of the TermReplacements here in this function.  You probably want `freeVariables(es).toList` here.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,50 @@
...
+import at.logic.gapt.proofs.{ Context, MutableContext, SequentConnector }
+
+object CreateASchemaVersion extends LKVisitor[MutableContext] {
+  override protected def recurse( p: LKProof, ctx: MutableContext ): ( LKProof, SequentConnector ) = {
+    val newNames = ctx.newNameGenerator
+    p match {
+      case proof @ InductionRule( casesI, form, typeTerm ) =>
+        val formNorm = BetaReduction.betaNormalize( form( typeTerm ) )
+        val newVarForDef = rename( Var( "x", typeTerm.ty ), freeVariables( proof.conclusion ) )
+        val es = proof.endSequent.updated( proof.mainIndices.head, BetaReduction.betaNormalize( form( newVarForDef ).asInstanceOf[Formula] ) )
```
gebner: Instead of `formNorm` you probably want to use `BetaReduction.betaNormalize( form( newVarForDef ).asInstanceOf[Formula]` everywhere (make a val for it though).  This should solve the substitution problems.

core/src/main/scala/at/logic/gapt/proofs/lk/NInductionToSchema.scala
```diff
@@ -0,0 +1,50 @@
...
+              ContractionMacroRule( CutRule( ProofLink( syntacticMatching( formNorm, subproof.endSequent( hypoth ) ).get( proofName ), outputSeq ), outputSeq.indexOf( outputProof.endSequent( hypoth ) ), outputProof, hypoth ), sigma( endSequentLeft ) )
+            } )
+/*            else {
+              val newante = endSequentLeft.antecedent.filter( t => subproof.endSequent.indexOfOption( t ).isEmpty && ( !t.contains( typeTerm ) || !freeVariables( t ).contains( typeTerm.asInstanceOf[Var] ) ) )
+              val newsuc = endSequentLeft.succedent.filterNot( t => subproof.endSequent.indexOfOption( t ).isEmpty && ( !t.contains( typeTerm ) || !freeVariables( t ).contains( typeTerm.asInstanceOf[Var] ) ) )
+              WeakeningRightMacroRule( WeakeningLeftMacroRule( subproof, newante ), newsuc )
+            }*/
+            ArithmeticInductionToSchema( finProof, sigma( proofName ) )( ctx )
+        }
+        withIdentitySequentConnector( ProofLink( TermReplacement( proofName, newVarForDef, typeTerm ), proof.endSequent ) )
```
gebner: Get rid of the TermReplacement.

tests/src/test/scala/at/logic/gapt/prooftool/ProoftoolViewableTest.scala
```diff
@@ -30,7 +30,7 @@ class ProoftoolViewableTest extends Specification {
     implicitly[ProoftoolViewable[ExtendedProofDatabase]]
 
     def forall_[Data] =
```
gebner: We can drop the "universal quantifier" now (just remove the def-line).

core/src/main/scala/at/logic/gapt/proofs/lk/instantiateProof.scala
```diff
@@ -12,32 +13,35 @@ object instantiateProof {
...
+  }
   def apply( proof: LKProof )( implicit ctx: Context ): LKProof =
     buildProof( proof, ctx )
 
   private object buildProof extends LKVisitor[Context] {
     override def visitProofLink( link: ProofLink, otherArg: Context ): ( LKProof, SequentConnector ) = {
       val ( connInstProof2Link, instProof ) = instantiateProof.withConnector( link.referencedProof )( otherArg )
-      ( instProof, connInstProof2Link )
+      val weakSeq = link.referencedSequent.filterNot( t => ( instProof.endSequent.find( w => t.equals( w ) ).nonEmpty ) )
+      val finProof = WeakeningRightMacroRule( WeakeningLeftMacroRule( instProof, weakSeq.antecedent ), weakSeq.succedent )
```
gebner: Most methods in gapt take sequents as arguments, there's rarely any need to decompose a sequent into its antecedent and succedent.  Here you can compute finProof in a single call to WeakeningMacroRule:
```scala
val finProof = WeakeningMacroRule(instProof, link.referencedSequent)
```

Issue #646, opened 2017-09-08
None of the TIP benchmarks work with the current tip-tools


gebner:
Yes, the TIP format changed, again.  The main problem are the 96 TIP files used in the tests, none of them can be read by the current tip-tools.

The breaking format change is that `(assert-not)` is called `(prove)` now.  That is, there is no way we can support both tip-tools versions.

For the next release, we will switch to the tip-tools master, and update the tests.  The old tip-tools version will then no longer work with gapt.

gebner:
BTW, the user manual is still failing.

PR #657, opened 2017-10-23
Tip examples

jannikvierling:


gebner:
So to summarize, we currently parse TIP files in the following 4 steps:
 1. Run the tip tool on it. (For monomorphization and other stuff)
 2. Parse the resulting TIP file into an `SExpression` (`SExpressionParser`).
 3. Convert the `SExpression` into a `TIPProblem` (`TipSmtParser`).
 4. Convert the `TIPProblem` into a Context plus a goal (`TipProblem.ctx`, `TipProblem.toSequent`).

With the change in this PR, we would skip the first three steps for the TIP example proofs.  Each TIP example proof constructs a `TIPProblem` by hand, and then uses step 4 to convert it into a Context plus a goal to prove.  This fixes the immediate bug we have in issue #646: namely that the current TIP tool cannot parse the TIP files included in GAPT.

But I don't think it helps us much in the long run.  I think we should take this opportunity to also get rid of step 4 in the TIP example proofs.
  * We can then change the TIP import without breaking all the 65 example proofs.  Right now we can't even change the *order* of the formulas in the output without breaking them...  (This PR does not address this problem.)
  * We can get rid of hypothesis names like `h17` and replace them by something human-readable.

jannikvierling:
If I understand well you say the formulas in the resulting sequent should be named manually for each example so that the formulas are not tied to their ordering in the definition and the names are more human-readable. Is that right? I am then not sure if we should keep the definition files so that we can reuse the problem definitions elsewhere or if we should integrate the problem definition into the proofs files.

gebner:
> If I understand well you say the formulas in the resulting sequent should be named manually for each example so that the formulas are not tied to their ordering in the definition and the names are more human-readable. Is that right?

Yes, that's what I meant.

> I am then not sure if we should keep the definition files so that we can reuse the problem definitions elsewhere or if we should integrate the problem definition into the proofs files.

My preference would be to have them inline, like here for example:
https://github.com/gapt/gapt/blob/9fead5e9619846df8ab5c045e712cc48c1fb5531/examples/tip/prod/prop_01.scala#L10-L22

We can still use them from other files by referencing `prop_01.ctx` and `prop_01.sequent`, although I don't see where that would happen.

gebner:
👍 I'm impressed. That's fast, and exactly how I think it should be.

gebner:
@jannikvierling Do we also need to update the `testing/TIP/tip-benchmarks-0.2.tar.gz` file after this change?

jannikvierling:
@gebner You're right I forgot about this file. There are also some examples in the user manual which use the TIP files and I don't yet know what I will do with these examples.

PR #658, opened 2017-10-25
add quoted atoms to s-expression parser

jannikvierling:


core/src/main/scala/at/logic/gapt/formats/lisp/SExpression.scala
```diff
@@ -66,8 +66,9 @@ class SExpressionParser( val input: ParserInput ) extends Parser {
 
   def Str = rule { '"' ~ capture( zeroOrMore( noneOf( "\"" ) ) ) ~ '"' ~ WhiteSpace ~> lisp.LAtom }
   def Atom = rule { capture( oneOrMore( noneOf( "() \n\r\t\f;\"" ) ) ) ~ WhiteSpace ~> lisp.LAtom }
+  def QuotedAtom = rule { "|" ~ capture( oneOrMore( noneOf( "\\|" ) ) ) ~ "|" ~ WhiteSpace ~> lisp.LAtom }
```
gebner: So `|\||` is not a quoted atom?  [The "official" grammar allows it.](https://github.com/tip-org/tools/blob/master/tip-lib/src/TIP.bnfc#L117)

Not that it matters, I just got curious when you explicitly excluded the backslash.

jannikvierling: Ok, then I think we should allow it too.

gebner:
Why is this not merged yet?

gebner:
BTW, this PR broke the user manual again.

PR #660, opened 2017-10-31
update regression test tip files

jannikvierling:


gebner:
"Looks good to me."

(I haven't looked at it, and jenkins will hopefully complain if it's not good. 😄)

Issue #664, opened 2017-11-06
TIP import: :let attribute on define-fun-rec


gebner:
```
± viper tip2015/sort_nat_MSortBUCount.smt2
Exception in thread "main" scala.MatchError: (define-fun-rec map :let :source Prelude.map ((f fun1) (x list2)) list (match x (case nil2 nil) (case (cons2 y xs) (cons (apply1 f y) (map f xs)))))
```

PR #665, opened 2017-11-07
Tip smt parser

jannikvierling:
Fixes issue #664.

### DIFF
--- a/core/src/main/scala/at/logic/gapt/formats/lisp/SExpression.scala
+++ b/core/src/main/scala/at/logic/gapt/formats/lisp/SExpression.scala
@@ -14,7 +14,13 @@ import scala.util.{ Failure, Success, Try }
  */
 sealed abstract class SExpression
 
-case class LAtom( name: String ) extends SExpression {
+sealed abstract class LAtom extends SExpression
+
+case class LKeyword( name: String ) extends LAtom {
+  override def toString = s":$name"
+}
+
+case class LSymbol( name: String ) extends LAtom {
   override def toString = name
 }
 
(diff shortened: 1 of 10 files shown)


core/src/main/scala/at/logic/gapt/formats/lisp/SExpression.scala
```diff
@@ -14,7 +14,13 @@ import scala.util.{ Failure, Success, Try }
  */
 sealed abstract class SExpression
 
-case class LAtom( name: String ) extends SExpression {
+sealed abstract class LAtom extends SExpression
```
gebner: This distinction may require changes in the SMT solver interfaces as well.  We at least use s-expressions to write `(set-option :produce-unsat-cores true)`.  (AFAICT, this wouldn't even break since LSymbol fails to check that name doesn't begin with a colon.)

Is there a reason we can't just make an LKeyword matcher that matches LAtoms whose name starts with a colon?

jannikvierling: I think it is cleaner to distinguish between symbols and keywords. If we take keywords to be the symbols that begin with `:`, how would we handle the symbol `:`? 

That said, I think matcher for LKeywords may also work.

gebner: Ah, I see.  So we want to distinguish between `|:foo|` and `:foo`?  Then this distinction is reasonable.  But I think `LSymbol(":foo")` should print `|:foo|`, and we should change `Session.setOption` to use LKeyword.

PR #666, opened 2017-11-10
New schema example

Ermine516:
Just the tutorial concerning proof schema construction and an example proof schema.


gebner:
Please run `sbt format`.

gebner:
Also you should add `package at.logic.gapt.examples` to each of the tutorial files, and give each one a different object name like `FirstSchema0`, `FirstSchema1`, etc.

We compile all the example files to make sure that they continue to work, so they need to have different class names.

PR #668, opened 2017-12-18
Explicit induction

gebner:
@mschlaipfer I'm making a PR so that I can look and comment on the whole diff.


core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
     theorySolver: HOLClause => Option[LKProof] ) extends SolveUtils {
   type Error = ( Seq[ETImp], ExpansionSequent )
 
-  def apply( expansionProof: ExpansionProof )( implicit ctx: Context = Context() ): UnprovableOrLKProof =
-    solve( expansionProof.cuts, expansionProof.nonCutPart ).
-      map( WeakeningMacroRule( _, expansionProof.nonCutPart.shallow ) )
+  abstract class Case
+  case class StepCase( evs: Seq[Var], atoms: List[ETAtom] ) extends Case
```
gebner: I think the distinction between step and base cases is not very useful in this context.  Step cases can have 1,2,3,... induction hypotheses--base cases should just be step cases with 0 induction hypotheses.  Since we need to handle an arbitrary number of induction hypotheses anyhow, zero should just work as well.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
...
+  case class BaseCase( atom: ETAtom ) extends Case
```
gebner: Note that `ETAtom` is way too specific here.  The induction formula may be quantified, which will not be an atom.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
...
+    }
+    def getEvs( et: ExpansionTree ): List[Var] = {
+      et match {
+        case ETStrongQuantifier( _, ev, ch ) => ev :: getEvs( ch )
+        case _                               => List.empty
+      }
+    }
+    def toCase( et: ExpansionTree ): Case = {
+      et match {
+        case q @ ETStrongQuantifier( _, _, ch ) => StepCase( getEvs( q ), getAtoms( ch ) )
```
gebner: As mentioned above, both `getEvs` and `getAtoms` will fail in the case of quantified induction formulas.  I would compute the number of expected induction hypotheses and their number of eigenvariables here, and unwrap the implication and conjunctions only as far as necessary.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
...
+    }
+
+    val inductions = for {
+      inductionAxiomExpansion <- expansionProof.expansionSequent.antecedent
+      if indAxioms.contains( inductionAxiomExpansion.shallow )
+      sequent <- inductionAxiomExpansion( HOLPosition( 1 ) )
+      hyps <- sequent( HOLPosition( 1 ) )
+      suc <- sequent( HOLPosition( 2 ) )
+      cs = indAxioms( inductionAxiomExpansion.shallow )
+    } yield Induction( cs.zip( hyps.immediateSubProofs.map( toCase( _ ) ) ), hyps, suc.asInstanceOf[ETWeakQuantifier] )
```
gebner: Note: `suc` may be something else than an `ETWeakQuantifier`, it could also be a `ETWeakening`.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
...
+      orElse( tryUnary( theory, expSeq ) ).
+      orElse( tryCut( theory, expSeq ) ).
+      orElse( tryInduction( theory, expSeq ) ).
+      orElse( tryBinary( theory, expSeq ) ).
+      orElse( tryTheory( theory, expSeq ) ).
+      getOrElse( Left( ( theory.getExpansionTrees ) -> expSeq ) ).
       map {
         ContractionMacroRule( _ ).
-          ensuring { _.conclusion isSubsetOf expSeq.map { _.shallow } }
+          ensuring { _.conclusion isSubsetOf expSeq.shallow }
```
gebner: 👍

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +16,144 @@ class ExpansionProofToLK(
...
       }
   }
 
-  private def tryAxiom( cuts: Seq[ETImp], expSeq: ExpansionSequent ): Option[UnprovableOrLKProof] = {
+  private def tryAxiom( theory: Theory, expSeq: ExpansionSequent )( implicit ctx: Context ): Option[UnprovableOrLKProof] = {
```
gebner: This function does not require a context argument.  (AFAICT, none of them do.)

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +214,77 @@ class ExpansionProofToLK(
...
+          else solve( Theory( newCuts, theory.inductions ), cut2 +: expSeq ) map { p2 =>
             if ( !p2.conclusion.contains( cut2.shallow, Polarity.InAntecedent ) ) p2
             else CutRule( p1, p2, cut1.shallow )
           }
         }
     }
   }
 
+  private def tryInduction( theory: Theory, expSeq: ExpansionSequent )( implicit ctx: Context ): Option[UnprovableOrLKProof] = {
+    val upcomingEVs = ( for {
```
gebner: Please make this a `lazy val` so that it is not computed if there are no inductions.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +214,77 @@ class ExpansionProofToLK(
...
+      et <- theory.getExpansionTrees ++ expSeq.elements
+      ETStrongQuantifier( _, ev, _ ) <- et.subProofs
+    } yield ev ).toSet
+
+    theory.inductions.zipWithIndex.collectFirst {
+      case ( Induction( constructorsSteps, hyps, suc ), i ) if freeVariables( hyps.shallow ) intersect upcomingEVs isEmpty =>
+
+        val newInductions = theory.inductions.zipWithIndex.filter { _._2 != i }.map { _._1 }
+
+        def recSteps( constructorsSteps: Seq[( Const, Case )], cases: Seq[InductionCase] ): UnprovableOrLKProof = {
```
gebner: In case you feel adventurous and want to do "hard-core" functional programming, you can check out [traverse](http://eed3si9n.com/herding-cats/Traverse.html).  Essentially, in this case here traverse takes a function `(Const, Case) => Either[Error, LKProof]` and returns a function `List[(Const,Case)] => Either[Error, List[LKProof]]` (recall that `UnprovableOrLKProof` is defined as `Either[Error, LKProof]`).

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +214,77 @@ class ExpansionProofToLK(
...
+                if ( !p.conclusion.contains( atom.shallow, Polarity.InSuccedent ) ) Right( p )
+                else {
+                  val sIdx = p.conclusion.indexOf( atom.shallow, Polarity.InSuccedent )
+                  recSteps( tail, InductionCase( p, c, Seq.empty, Seq.empty, sIdx ) +: cases )
+                }
+              }
+            case ( c, StepCase( evs, atoms ) ) +: tail =>
+              val ( ant, suc ) = atoms.splitAt( atoms.length - 1 )
+              solve( Theory( theory.cuts, newInductions ), ant ++: expSeq :++ suc ) flatMap { p =>
+                if ( !( ant.forall( a => p.conclusion.contains( a.shallow, Polarity.InAntecedent ) )
```
gebner: You need two distinguish *three* cases here:
(1) *None* of expected formulas (induction hypotheses, conclusion) are present: then we can just return p.
(2a) *Some* of the expected formulas are present, others not.  In this case you need to add the missing formulas via weakening and continue with (2b).  You *cannot* return p, since its conclusion is not a subset of the shallow sequent of `expSeq`.
(2b) *All* of the expected formulas are present.  Construct the induction case.

Practically you'd implement (2a) and (2b) as the same case, you just add a call to `WeakeningMacroRule` at the top.

core/src/main/scala/at/logic/gapt/proofs/lk/LKToExpansionProof.scala
```diff
@@ -150,5 +148,9 @@ object LKToExpansionProof {
       val ( subCuts, subSequent ) = extract( subProof )
 
       ( subCuts, subSequent.delete( aux ) :+ ETDefinition( main, subSequent( aux ) ) )
+
+    case p @ InductionRule( _, _, _ ) =>
+      val ( subCuts, subSequent ) = extract( makeInductionExplicit( p ) )
```
gebner: Note: `makeInductionExplicit` traverses the proof recursively.  It will convert *all* induction rules.  In particular it may add more than one induction axiom, so just taking `Ant(0)` will not be enough.

I think we should not make a call to makeInductionExplicit here and eventually construct the expansion trees directly here.  But for now, it's fine.  You could even call makeInductionExplicit right before AtomicExpansion for now.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
mschlaipfer: Why only the numbers? What do you consider "as far as necessary"?

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
mschlaipfer: Not sure I follow this. `InductionCase` is not an `LKProof`.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
gebner: In the for-loop below, hyps is an expansion tree with the shallow formula `!ev1_1... (IH1_1 -> IH1_2 -> ... -> A1) & !ev2_1... (IH2_1 -> ... -> A2) & .... & !evn_1... (IHn_1 -> ... -> An)`.  In order to decompose this expansion tree, you just need to know 1) how many constructors there are (to decompose the outer conjunction), 2) how many eigenvariables there are (to decompose the quantifiers), and 3) how many induction hypotheses there are (to decompose the implications).

By "only as far as necessary" I mean that you shouldn't try to look at the inner structure of `A1`, for example.  In particular, you can't decompose `IH1_1 -> .... -> A1` by removing implications as long as possible---this may by incorrect since A1 may be an implication.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
gebner: Good catch!  I meant `InductionCase`:

Essentially, in this case here traverse takes a function `(Const, Case) => Either[Error, InductionCase]` and returns a function `List[(Const,Case)] => Either[Error, List[InductionCase]]`.

mschlaipfer: So what if none of the formulas are present (see your next point)? I guess you want to use `traverse` to turn a list of `Either`s into an `Either` of a list? I don't get it yet.

gebner: Ah, yes.  `traverse` might not be the best solution then.  The `recSteps` function is probably the best  solution since it has the right backtracking behavior.

For reference, you'd call traverse as `constructorSteps.toList.traverse(f)`, where `f` is the function that handles a single Case.  If `f` fails for a single element (returns `Left(...)`), then traverse also returns `Left(...)`.  If `f` returns `Right(c1), ..., Right(cn)` for each of the elements of `constructorSteps` respectively, then traverse returns `Right(List(c1, ..., cn))`.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +215,78 @@ class ExpansionProofToLK(
...
+                    suc.collect { case s if !p.conclusion.contains( s.shallow, Polarity.InSuccedent ) => s.shallow } )
+                  val aIdxs = ant.map( a => pWkn.conclusion.indexOf( a.shallow, Polarity.InAntecedent ) )
+                  val sIdx = pWkn.conclusion.indexOf( suc.head.shallow, Polarity.InSuccedent )
+                  recSteps( tail, InductionCase( pWkn, c, aIdxs, evs, sIdx ) +: cases )
+                }
+              }
+            case Nil =>
+              val App( _, qfFormula: Abs ) = suc.shallow
+
+              val expr = suc match {
```
gebner: A single induction axiom may have been extracted from multiple induction rules---basically, one induction rule per weak quantifier instance in `suc`.  So we probably need to do something more clever here than 1 induction + 1 cut.  I see two options:
* n inductions + n cuts.  This is essentially the current code, just iterated once for each weak quantifier term.  We use one induction rule for each weak quantifier instance, followed by a cut.  I'm pretty sure this gets us into trouble with the dependency order on the eigenvariables though.  Not recommended.
* 1 induction + 1 forall-right + 1 cut + rest handled by solve.  Let `φ(x)` be the induction formula.  Then `suc` is an expansion tree with the shallow formula `∀x φ(x)` and in-antecedent polarity.  And `solve(newTheory, suc +: expSeq)` returns an expansion sequent with shallow sequent `∀x φ(x), ... :- ...`.  We then create the induction rule with a fresh variable as term (just distinct from the shallow sequent), add a forall-right rule and combine it with the other proof using a cut.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -15,93 +17,144 @@ class ExpansionProofToLK(
...
+  case class Case( evs: Seq[Var], step: Seq[ExpansionTree] )
+  case class Induction( constructorsSteps: Seq[( Const, Case )], hyps: ExpansionTree, suc: ExpansionTree )
+  case class Theory( cuts: Seq[ETImp], inductions: Seq[Induction] ) {
+    def getExpansionTrees: Seq[ETImp] = cuts ++ inductions.map( i => ETImp( i.hyps, i.suc ) )
+  }
+
+  def apply( expansionProof: ExpansionProof )( implicit ctx: Context = Context() ): UnprovableOrLKProof = {
 
-  private def solve( cuts: Seq[ETImp], expSeq: ExpansionSequent ): UnprovableOrLKProof = {
+    val indAxioms = ctx.get[StructurallyInductiveTypes].constructors.map {
```
gebner: Eventually, it might make sense to move this induction-axiom parsing logic into a separate object, say `ETInduction`, just like we have `ETCut` now for cuts.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
shetzl: Yes, you are right, it is more complicated.

I would prefer the 2nd variant.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +173,70 @@ class ExpansionProofToLK(
...
 
+  private def tryInduction( theory: Theory, expSeq: ExpansionSequent ): Option[UnprovableOrLKProof] = {
+    lazy val upcomingEVs = ( for {
+      et <- theory.getExpansionTrees ++ expSeq.elements
+      ETStrongQuantifier( _, ev, _ ) <- et.subProofs
+    } yield ev ).toSet
+
+    theory.inductions.zipWithIndex.collectFirst {
+      case ( ETInduction.Induction( constructorsSteps, hyps, suc ), i ) if freeVariables( hyps.shallow ) intersect upcomingEVs isEmpty =>
+
```
gebner: Please remove the empty line here.

core/src/main/scala/at/logic/gapt/proofs/expansion/ExpansionProofToLK.scala
```diff
@@ -154,27 +173,70 @@ class ExpansionProofToLK(
...
+        val newInductions = theory.inductions.zipWithIndex.filter { _._2 != i }.map { _._1 }
+
+        def recSteps( constructorsSteps: Seq[( Const, ETInduction.Case )], cases: Seq[InductionCase] ): UnprovableOrLKProof = {
+          constructorsSteps match {
+            case ( c, ETInduction.Case( evs, steps ) ) +: tail =>
+              val ( ant, suc ) = steps.splitAt( steps.length - 1 )
```
gebner: I think `steps` should be an ExpansionSequent.  Then we can write the following without treating antecedent and succedent separately:
 * `solve(newTheory, expSeq ++ steps)`
 * `p.conclusion.intersect(steps.shallow).isEmpty`
 * `WeakeningMacroRule(p, steps.shallow, strict=false)`

Unrelated to the type: the name `steps` feels a bit misleading to me.  In the context of induction, in the natural numbers for example I would expect steps to refer to the step case---that is, the case for the constructor `s`.  However here, steps actually refers to the induction hypotheses plus the formula in the succedent.  Maybe `auxiliary` would be a better name?

mschlaipfer: That's much better, good suggestion. I'll also rename it.

PR #677, opened 2018-01-14
Replace superset symbol by function arrow

gebner:
I'm of course talking about the implication arrow.
 * Most of the current gapt developers seem to prefer this notation.
 * It increases consistency with the ascii notations `->` and `<->`, as well as the scala function `-->`.
 * We already used → in comments.


gebner:
It would be the first configuration option that we have.  I'm not even sure where to put it---environment variables, configuration files?

In general, I'm partial to the "there is only way way to do it" principle.  If there is no good reason to have two ways to do something, we should remove one of them.  The fewer options we have, the fewer untested and broken corner cases there are.  (One test broke because the name of the implication constant changed...)  Particularly when it comes to syntax, consistency is valuable for its own sake as well---in my experience it makes it easier to tune out the formal noise and  see just the actual content.

If there is a really big backlash against this change, I'd rather keep the current notation.

gebner:
@loewenheim Just to count votes: 1) do you like ⊃, 2) do you like →, or 3) are you indifferent?

gebner:
> Do you think there's a case for making this user-configurable?

With the latest changes in babel, we can make this happen.
```scala
ctx += Notation.Infix( "⊃", ImpC.name, Precedence.impl, leftAssociative = false )
```

Well, only kinda.  This only helps you if you have a Context that you can modify, and doesn't affect prooftool, etc.

Issue #669, opened 2017-12-21
specs2 ignores exceptions


gebner:
Exceptions that do not inherit from `java.lang.Exception` seem to be ignored by specs2 and do not show up in the test results.

gebner:
Bug report in specs2: https://github.com/etorreborre/specs2/issues/714

jannikvierling:
It seems to me that this is now fixed.

PR #685, opened 2018-02-08
Modified realizability

gebner:
(As usual, I'm just making a PR so that I can see all changes at a glance and add review comments.)


core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+
+object MRealizability {
+
+  def addRecursors( implicit ctx: Context ): Context = {
+    var ctxx = Context.empty
+
+    for ( ( name, constructors ) <- ctx.get[StructurallyInductiveTypes].constructors.filter( _._1 != "o" ) ) {
+      val typ = ctx.get[BaseTypes].baseTypes( name )
+      ctxx += InductiveType( typ, constructors )
+      val argTypes = constructors.map( x => x -> FunctionType.unapply( x.ty ).get._2 ) toMap
```
gebner: You can avoid calling `unapply` explicitly if you use a pattern-matching function as argument to `map`:
```scala
constructors.map{ case Const(_, FunctionType(_, ctrArgTys)) => ctrArgTys }.toMap
```
(Then I don't have to figure out what `_2` refers to.)

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+      val resultVariable = TVar( new NameGenerator( typeVariables( typ ) map ( _.name ) ).fresh( "a" ) )
+      val ngTermVariableNames = new NameGenerator( ctx.constants map ( _.name ) )
+
+      val constrVars = constructors.map( x => x -> Var(
+        ngTermVariableNames.fresh( "x" ),
+        argTypes( x ).foldRight( resultVariable: Ty )( ( y, z ) => if ( y == typ ) typ ->: resultVariable ->: z else y ->: z ) ) ) toMap
```
gebner: I really don't like foldLeft/foldRight.  For example here, I can't tell at a glance whether the order of the argument types is reversed or not.  How about the following?
```scala
FunctionType(resultVariable, argTypes(x).flatMap(argTy =>
  if (argTy == typ) Seq(typ, resultVariable) else Seq(typ)))
```

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+
+      val recursortype = constructors.foldRight( typ ->: resultVariable: Ty )( ( x, y ) => constrVars( x ).ty ->: y )
```
gebner:
```scala
FunctionType(resultVariable, constructors.map(constrVars(_).ty))
```

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+      val recursor = Const( name + "Rec", recursortype, typeVariables( recursortype ).toList )
```
gebner: A word of warning: `typeVariables(recursortype).toList` returns the type variables in a random (but deterministic) order.  This makes it hard to construct recursor terms afterwards (since you need to guess the order of the parameters).

I'd suggest to use `typ.params :+ resultVariable` instead.  Then you can construct recursors afterwards by just appending the result (motive) type to the parameters of the inductive type in the induction inference.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+      val argVars = argTypes.map( x => x._1 -> x._2.map( y => Var( ngTermVariableNames.fresh( "x" ), y ) ) )
+
+      val equations = constructors.map( x =>
+        (
+          App( recursor, constrVars.values.toVector :+ App( x, argVars( x ) ) ),
```
gebner: In case you find it more readable, you can also use function application instead of App--i.e., write `recursor(constructors.map(constrVars) :+ x(argVars))` instead.

Please note again that `constrVars.values.toVector` returns the values of `constrVars` in a random order.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+          argVars( x ).foldLeft( constrVars( x ): Expr )( ( y, z ) => if ( z.ty == typ ) App( App( y, z ), App( recursor, constrVars.values.toVector :+ z ) ) else App( y, z ) ) ) )
```
gebner: I find this line virtually impossible to decipher.  Please consider using flatMap and App instead of foldLeft.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+    systemT += InductiveType(
+      ty"1",
+      hoc"i : 1" )
+
+    mrealizeCases( proof )( systemT )
+  }
+
+  def mrealizeCases( proof: NDProof )( implicit ctx: Context ): Expr = {
+
+    normalize( proof match {
```
gebner: I don't think we should normalize the realizers by default.  This seems like it could be *really* expensive.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+        Abs(
+          freeVariables( proof.conclusion ).toSeq ++ variablesAntConclusion( proof ),
+          App(
+            App( mrealizeCases( leftSubProof ), freeVariables( leftSubProof.conclusion ).toSeq ++ variablesAntPremise( proof, 0 ) ),
+            App( mrealizeCases( rightSubProof ), freeVariables( rightSubProof.conclusion ).toSeq ++ variablesAntPremise( proof, 1 ) ) ) )
+
+      case ImpIntroRule( subProof, aux ) =>
+        val extraVar = Var( "z", flat( subProof.conclusion( aux ) ) )
+        Abs(
+          freeVariables( proof.conclusion ).toSeq ++ variablesAntConclusion( proof ),
```
gebner: You can combine both `Abs` calls into a single one if you prefer:
```scala
Abs(freeVariables(proof.conclusion).toSeq ++ variablesAntConclusion(proof) :+ extraVar, ...)
```

Note once again that `freeVariables(...).toSeq` returns the variables in a random order.  Generally, the order of elements in a `Map` or `Set` is unspecified (as they're implemented by a hash map in general).

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+      case TheoryAxiom( mainFormula ) =>
+        throw new MRealizerCreationException( proof.longName, "Not implemented yet." )
+
+      case EqualityElimRule( leftSubProof, rightSubProof, formulaA, variablex ) =>
+        Abs(
+          freeVariables( proof.conclusion ).toSeq ++ variablesAntConclusion( proof ),
+          App( mrealizeCases( rightSubProof ), freeVariables( rightSubProof.conclusion ).toSeq ++ variablesAntPremise( proof, 1 ) ) )
+
+      case EqualityIntroRule( term ) =>
+        Abs( freeVariables( proof.conclusion ).toSeq, le"i" )
```
gebner: After thinking a bit about it, I don't believe it is a good idea to provide closed realizers: (closed = no free variables, open = can contain free variables)
 1) We get lots of useless beta-redexes.  You have added a call to normalize to clean them up, but this has the unfortunate consequence of reducing the recursors as well.
 2) The order of the free variables is unspecified, and hence the order of the arguments of the realizers is random as well.
 3) From a code size point of view, there are way too many Abs and App calls here.

How about producing open realizers instead?  That is, the free variables of the realizer are the free variables of the proof.  There should be no need to add binders for the free variables.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,219 @@
...
+    variablesAntConclusionWithIndex( proof ).map( _._2 )
+  }
+
+  def variablesAntPremise( proof: NDProof, premiseNumber: Int ): Vector[Var] = {
+    val positions = proof.occConnectors( premiseNumber ).childrenSequent.antecedent.flatten
+    positions.flatMap( variablesAntConclusionWithIndex( proof ).toMap get )
+  }
+
+  private def variablesAntConclusionWithIndex( proof: NDProof ): Vector[( SequentIndex, Var )] =
+    proof.conclusion.zipWithIndex.antecedent.map( x => ( x._2, Var( s"x${x._2.toInt}", flat( x._1 ) ) ) )
```
gebner: This is gonna clash with the variables in the proof.

My suggestion would be to add an extra `vars: Sequent[Var]` argument to `mrealizeCases` that contains the variables for the realizers of the assumptions.  You can move them upwards to the i-th premise of the inference by using `proof.getSequentConnector(i).parent(vars)`.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
shetzl: I agree. Moreover, one may even want to look at an unnormalized realizer in order to better understand how it relates to the proof.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,316 @@
...
+
+import at.logic.gapt.expr.{ App, typeVariables, Ty, _ }
+import at.logic.gapt.proofs.Context.{ BaseTypes, InductiveType, PrimRecFun, StructurallyInductiveTypes }
+import at.logic.gapt.proofs._
+import at.logic.gapt.proofs.nd._
+import at.logic.gapt.utils.NameGenerator
+
+object MRealizability {
+
+  implicit var systemT: Context = Context.empty
```
gebner: This is a mutable global variable.  Please don't do that--it makes it hard to reason about the code, and it is unsafe in the presence of multi-threading.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,316 @@
...
+      case ExcludedMiddleRule( leftSubProof, aux1, rightSubProof, aux2 ) =>
+        throw new MRealizerCreationException( proof.longName, "This rule is not admitted in Heyting Arithmetic." )
+    } )
+  }
+
+  def flat( formula: Formula ): Ty = formula match {
+    case Bottom()                         => ty"1"
+    case Top()                            => flat( Imp( Bottom(), Bottom() ) )
+    case Eq( _, _ )                       => ty"1"
+    case Atom( _, _ )                     => ty"1" // ?
```
gebner: This `?` is actually an interesting consideration: we might want to interpret certain atoms as non-unit types.

For example, consider the divisibility predicate `dvd` where `dvd x y` means that `x` divides `y`.  Think of `dvd x y` being defined as `∃z y = x*z`.  The natural interpretation for `dvd x y` would then be the `nat` type (the same as the definition), and not unit.

MarielleLouise: Hey Gabriel, I only notice now that you reviewed all the code that I comitted lately, thank you for that!! Sorry if it seemed that I have been ignoring all this, last time you made comments to it, I received an email from github about it, but now I didnt - so I wasn't aware of all this. I will consider your comments and make changes to the code accordingly very soon :).

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
MarielleLouise: I agree and I changed this with the commit I just made: allow the realizer to contain free variables that occur in the formula it realizes.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
MarielleLouise: I normalized by default because the abstraction and application of free variables and formulas in the antecedent create a lot of redexes. The free variables redexes are not there anymore, but those for the formulas in the antecedent are. I am thinking to do that differently though. Right now I create a realizer for a formula C1 -> ... -> Ck -> A, when realizing a sequent C1, ..., Ck :- A. But I think that it maybe is a better idea to instead create a realizer for the formula A, with free variables of the type flat(Ci) as placeholders for realizers for C1, ..., Ck. 

I decided to go for only beta-reduction now, so at least it leaves the recursors as they are.

gebner: @MarielleLouise Creating realizers of type flat(A) is a good approach.

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,283 @@
...
+      case AndIntroRule( leftSubproof, rightSubproof ) =>
+        Abs(
+          variablesAntConclusion( proof ),
+          le"pair(${App( mrealizeCases( leftSubproof ), variablesAntPremise( proof, 0 ) )},${App( mrealizeCases( rightSubproof ), variablesAntPremise( proof, 1 ) )})" )
+
+      case OrElimRule( leftSubProof, middleSubProof, aux1, rightSubProof, aux2 ) =>
+        val realizerAOrB = App( mrealizeCases( leftSubProof ), variablesAntPremise( proof, 0 ) )
+        Abs(
+          variablesAntConclusion( proof ),
+          le"natRec(${
```
gebner: Why is this a recursor on natural numbers?  Shouldn't this be a recursor on an inductive `sum` type?

MarielleLouise: I use here that he disjunction is definable in arithmetic by  \exists x (((x=0) \rightarrow A) \land ( \neg (x=0) \rightarrow B )). But it is not the right thing to do because it depends on the inductive type of natural numbers, so I need another more general solution for this. I am not familiar with the sum type, but Matthias also mentioned it and I think it would be good if I will use this as well for the disjunction!

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
MarielleLouise: It has been some while since you have made these and other comments about the code concerning adding recursors to the context, but I have meanwhile improved the readability of the code, and your suggestions were very helpful, thank you!

MarielleLouise:
I think I have everything done for modified realizability. Let me kwow if something is missing or should be changed :).

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,402 @@
...
+        App( mrealizeCases( subProof, varsAntPrem( proof, variables, 0 ), ng ), term )
+
+      case ExistsIntroRule( subProof, formula, term, variable ) =>
+        le"pair($term,${mrealizeCases( subProof, varsAntPrem( proof, variables, 0 ), ng )})"
+
+      case ExistsElimRule( leftSubProof, rightSubProof, aux, eigenVariable ) =>
+        val mrealizerLeft = mrealizeCases( leftSubProof, varsAntPrem( proof, variables, 0 ), ng )
+        val sub1 = Substitution( eigenVariable, le"pi1($mrealizerLeft)" )
+        val extraVar = Var( ng.fresh( "y" ), flat( rightSubProof.conclusion( aux ) ) )
+        val sub2 = Substitution( extraVar, le"pi2($mrealizerLeft)" )
```
gebner: You can combine `sub1` and `sub2` into a single substitution:
```scala
val sub = Substitution(
  eigenVariable -> le"pi1($mrealizerLeft)",
  extraVar -> le"pi2($mrealizerLeft)")
```

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
```diff
@@ -0,0 +1,402 @@
...
+      case _ => typee
+    }
+  }
+
+  def remEmpProgTypes( types: List[Ty] )( implicit systemT: Context ): List[Ty] =
+    types.map( remEmpProgType( _ ) )
+
+  // Given a proof and variables for the formulas in the antecedent of the conlusion,
+  // returns those variables that occur as well in the antecedent of the conclusion of the premise at premisenumber.
+  private def varsAntPrem( proof: NDProof, varsAntConcl: Map[SequentIndex, Var], premiseNumber: Int ): Map[SequentIndex, Var] = {
```
gebner:
```scala
for {
  (i, v) <- varsAntConcl
  j <- proof.occConnectors(premiseNumber).parentOption(i)
} yield j -> v
```

doc/user_manual.tex
```diff
@@ -1460,7 +1460,33 @@ \section{Expansion proofs}
 can selectively expand quantifiers by clicking on them,
 see~\cite{Hetzl13Understanding} for a detailed description.
 
+\section{Modified realizability}
 
+The GAPT-system contains an implementation of modified realizability. This is a variant of realizability where the realizers are terms of system T: an extension of the simply typed lambda calculus with inductive types and recursors that allow for recursive computation over terms of those types. 
+
+The implementation extracts realizers for theorems from their proofs in the natural deduction calculus. If the conclusion of the proof depends on a number of hypothesis, i.e. there are formulas $A_1,...,A_n$ in the antecedent of the final sequent, the algorithm returns a term $M$ with free variables  $x_1,...,x_n$ such that $M[M_1/x_1]...[M_n/x_n]$ is a realizer for the succedent, where $M_i$ realizes $A_i$. The formulas in the conclusion are also allowed to have free variables. In this case, the substitution of free variables in both the formulas and its realizer by the same closed terms, gives a realizer for the resulting sentence.
```
gebner: You should mention that this only works for first-order proofs.

doc/user_manual.tex
```diff
@@ -1460,7 +1460,33 @@ \section{Expansion proofs}
...
+
+The natural deduction calculus contains two rules that are problematic for the extraction of realizers, namely the theorem axiom, which can be instantiated by any formula, and a rule for the lemma of the excluded middle. When these are used, a realizer can not be computed - it might not even exists. In the case of the theory axiom, we would need a realizer for the theorem that is introduced by the axiom, and for the excluded middle rule, we would need a realizer for $A \lor \neg A$ to compute the realizer for the conclusion of the rule. These rules therefore introduce free variables in the resulting realizers, that serve as placeholders for the realizers we need.
```
gebner: *theory* axiom

gebner: use `---` instead of ` - `

doc/user_manual.tex
```diff
@@ -1460,7 +1460,33 @@ \section{Expansion proofs}
...
+
+Consider the following proof, that is made after defining up an implicit context containing the natural numbers and the primitive recursive function `+` :
```
gebner: use `$+$` or `\cli{+}` instead of `` `+` ``.  This is latex, not markdown.

doc/user_manual.tex
```diff
@@ -1460,7 +1460,33 @@ \section{Expansion proofs}
...
+
+The natural deduction calculus contains two rules that are problematic for the extraction of realizers, namely the theorem axiom, which can be instantiated by any formula, and a rule for the lemma of the excluded middle. When these are used, a realizer can not be computed - it might not even exists. In the case of the theory axiom, we would need a realizer for the theorem that is introduced by the axiom, and for the excluded middle rule, we would need a realizer for $A \lor \neg A$ to compute the realizer for the conclusion of the rule. These rules therefore introduce free variables in the resulting realizers, that serve as placeholders for the realizers we need.
+
+Consider the following proof, that is made after defining up an implicit context containing the natural numbers and the primitive recursive function `+` : 
+
+\begin{clilisting}
+gapt> val p4 = nd.ForallIntroRule(p3,hov"x:nat",hov"x:nat")
```
gebner: The examples in the user manual should be self contained.  Can you copy the definition of `p1, p2, p3` here from the ND section?

tests/src/test/scala/at/logic/gapt/proofs/nd/MRealizabilityTest.scala
```diff
@@ -0,0 +1,494 @@
...
+      val a4 = BottomElimRule( a3, hof"0=0" )
+      val b1 = LogicalAxiom( hof"0=0" )
+      val c1 = LogicalAxiom( hof"0=0" )
+      val c2 = LogicalAxiom( hof"¬0=0" )
+      val c3 = OrIntro1Rule( c1, hof"¬0=0" )
+      val c4 = OrIntro2Rule( c2, hof"0=0" )
+      val c5 = ExcludedMiddleRule( c3, Ant( 0 ), c4, Ant( 0 ) )
+      val a5 = OrElimRule( c5, b1, a4 )
+      val a6 = ImpIntroRule( a5 )
+      mrealizer( a6 ) must_== Abs(
```
gebner: Is there a reason you can't use the formula parser here?

gebner:
Also please merge master.

doc/user_manual.tex
gebner: use `\dots` instead of `...`

tests/src/test/scala/at/logic/gapt/proofs/nd/MRealizabilityTest.scala
MarielleLouise: No I guess I could just use it, with system T as context. For some reason I found it convenient to not use it at some point when makig examples, don't really remember why, and thought it wouldn't make a difference so didn't change it. But if there is a reason that I am not aware for changing it to using the parser, I could do that.

gebner: The only reason would be readability, this is just `le"^x matchSum(matchSum($r, ^y inl y, ^z inr z), ^w w, ^u y_4)"` if you set `r` to the realizer variable (which should probably get a nicer name as well).

BTW, what is the `y_4` doing here?

core/src/main/scala/at/logic/gapt/proofs/nd/m-realizability.scala
MarielleLouise: This doesn't work for the case of contraction, the contracted formula has two parents but `parentOption` gives only one. I made an extra example that doesn't work when using this (I called it contraction2). But I tried to take your idea and use `parents` instead, and it is much easier to read now so thanks :).

doc/user_manual.tex
MarielleLouise: I made a reference to the Appenndix for ND, is that also sufficient?

tests/src/test/scala/at/logic/gapt/proofs/nd/MRealizabilityTest.scala
MarielleLouise: I just wrote an extra line in the user manual to explain this sort of occurences of free variables in realizers.

doc/user_manual.tex
gebner: That's not what I meant.  I was talking about the `p3` in the code snippet here.  I think the snippet should also include the definitions for `p1`, `p2`, and `p3`:
```latex
\begin{clilisting}
gapt> val p1 = ...
gapt> val p2 = ...
gapt> val p3 = ...
gapt> val p4 = ...
...
```

Otherwise, the reader needs to look back a page or two to figure out what `p3` is.  Besides, it will break if we reorder the sections.  The goal of these code snippets is that you can take and copy-paste them to get a working example.  Having to collect code from different sections defeats this purpose.

shetzl: I agree. In fact, I tried looking back a page or two to figure out what p3 is, without success.

The examples in the user manual should allow someone who is curious about the feature to try it out by simple copy&paste. Therefore they need to be self-contained. If this means including a large formalisation of a proof, then so be it.

tests/src/test/scala/at/logic/gapt/proofs/nd/MRealizabilityTest.scala
gebner: I'm not sure I like it.  The classical approach is to define (ground) dummy terms by recursion on the type, e.g. `inl (pair i i) : sum (conj 1 1) 1`.

But I guess the variables work as well since they don't block reduction, so let's ship it...

gebner:
At the risk of repeating myself, please merge master.

doc/user_manual.tex
MarielleLouise: Sorry I misunderstood the comment, you are right, I will change it.

MarielleLouise:
Now that I have merged in the master, I am not able to compile anymore in sbt.. (sorry, I should not have committed it yet)  :

> [error] Symbol 'type gapt.proofs.gaptic.TacticCommands' is missing from the classpath.
[error] This symbol is required by 'package gapt.proofs.gaptic.package'.
[error] Make sure that type TacticCommands is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
[error] A full rebuild may help if 'package.class' was compiled against an incompatible version of gapt.proofs.gaptic.
[error] one error found
[error] (core / Compile / compileIncremental) Compilation failed
[error] Total time: 20 s, completed 6-apr-2018 14:56:33

I don't understand what that means unfortunately.. How can I try to solve this?

gebner:
That's a bug in sbt.

> A full rebuild may help if 'package.class' was compiled against an incompatible version of gapt.proofs.gaptic.

Just run `clean` and then `compile` should work again.

MarielleLouise:
Thanks, it works now!

MarielleLouise:
Sorry I was not being clear. The compilation works is what I meant, but I still needed to fix the tests. I fixed them now and comitted in the m-realizability branch. Sorry for being messy, I should be more clear next time..

gebner:
Oops I missed the failing test.  Apparently specs2 ignores some failing tests: #669.

tests/src/test/scala/at/logic/gapt/proofs/nd/MRealizabilityTest.scala
MarielleLouise: I could change it to that approach, i like it as well! I don't have much more time during the weekend but next week I could try to implement your suggestion.

examples/m-realizabilityExamples.scala
```diff
@@ -0,0 +1,222 @@
...
+import gapt.proofs.nd._
+import gapt.expr.{ Abs, TBase, _ }
+import gapt.proofs.{ Ant, Checkable, Context, Sequent, nd }
+import gapt.proofs.Context.{ InductiveType, PrimRecFun }
+
+object successor {
+
+  implicit var ctx = Context()
+
+  ctx += InductiveType(
```
gebner: All the examples fail now.

You're missing a declaration for the `+` operator here:
```scala
infix( "+", Precedence.plusMinus ); infix( "-", Precedence.plusMinus )
infix( "*", Precedence.timesDiv )
```

examples/m-realizabilityExamples.scala
```diff
@@ -0,0 +1,222 @@
+package gapt.examples
+
+import gapt.proofs.nd._
+import gapt.expr.{ Abs, TBase, _ }
+import gapt.proofs.{ Ant, Checkable, Context, Sequent, nd }
+import gapt.proofs.Context.{ InductiveType, PrimRecFun }
+
+object successor {
```
gebner: These should probably extend `TacticsProof`, then you can remove the `implicit var ctx`.

examples/m-realizabilityExamples.scala
MarielleLouise: Oops, yes, I fixed this in the tests, but forgot about the examples.

PR #691, opened 2018-02-20
LKt: proof terms for high performance cut normalization

gebner:
This PR introduces a syntactic variant of LK using proof terms, to support an efficient implementation of Gentzen-style cut-reduction.

As shown in the empirical evaluation section below, (depending on the proof) this implementation is 10<sup>4</sup>x faster than `ReductiveCutElimination`, 10<sup>2</sup>x faster than CERES, and 10x faster than expansion proof cut-elimination.

## Calculus

The proof system is modeled closely after the calculus described in the paper by Urban and Bierman [1].  The name LKt is new and refers to "*LK* with *t*erms".

Proofs in LKt operate on hypotheses (called names and co-names in [1]), which name formula occurrences in the current sequent.  We have untyped proof terms, and a typing judgment that tells us what sequent a proof proves.  Let us first give (a representative subset of) the syntax for the terms, where `Expr` refers to lambda expressions.  (In the present implementation, there are a few more terms representing inferences for equality, induction, definitions, proof links, and ⊤).
```
Hyp ::= -Nat | +Nat
Term ::= Ax(Hyp, Hyp)
       | Cut(Formula, Hyp: Term, Hyp: Term)
       | NegL(Hyp, Hyp: Term)
       | NegR(Hyp, Hyp: Term)
       | AndL(Hyp, Hyp: Hyp: Term)
       | AndR(Hyp, Hyp: Term, Hyp: Term)
       | AllL(Hyp, Expr, Hyp: Term)
       | AllR(Hyp, Var, Hyp: Term)
```

Notably, we do not have terms for weakening and contraction.  These are implicit: we can use the same hypothesis zero or multiple times.

A negative hypothesis refers to a formula in the antecedent, and a positive hypothesis refers to a formula in the succedent.  The notation `Hyp: Term` means that `Hyp` is bound in `Term`, c.f. the notation of abstract binding trees originating in [3].  This encoding of LK is also very similar to the encoding commonly used in logical frameworks (LF), see [4] for a description of such an approach.

Note that the proof terms only contain new information that is not contained in the end-sequent; only cut formulas, weak quantifier instance terms, and eigenvariables are stored.  We do not repeat the formulas or atoms of the end-sequent.

Let us now define the typing judgment.  We consider labelled sequents, which are set-sequents of hypothesis-formula pairs, such that 1) all hypotheses are distinct, and 2) hypotheses are positive iff they are in the succedent.  The judgment `π :: S` means that `π` proves the labelled sequent `S`.  We only show a few examplary typing rules here, the rest are analogous and correspond to the inferences rules in LK in a natural way.
```
-------------------------------------
 Ax(h₁, h₂)  ::  h₁: φ, Γ ⊢ Δ, h₂: φ


  π  ::  h₃: ψ, h₂: φ, h₁: φ ∧ ψ, Γ ⊢ Δ
-----------------------------------------
AndL(h₁, h₂: h₃: π)  ::  h₁: φ ∧ ψ, Γ ⊢ Δ


 π[x\y]  ::  h₂: φ(y), h₁: ∀x φ(x), Γ ⊢ Δ
------------------------------------------  (where y is not free in the sequent)
AllR(h₁, x, h₂: π)  ::  h₁: ∀x φ(x), Γ ⊢ Δ
```

We use the convention that new hypotheses "overwrite" old ones in labelled sequents, that is `h: φ, h: ψ, Γ ⊢ Δ` is equal to `h: φ, Γ ⊢ Δ`.  In particular, it is not necessary to rename hypotheses for type-checking.  In the present implementation, renaming of eigenvariables is avoided as well by keeping track of a substitution in addition to the sequent in the `LocalCtx` type.

## Cut-normalization

The cut-reduction procedure is very similar to the one presented in [1], so we only give a short overview as well as note important differences.

Rank reduction is implemented using so-called "proof substitutions", these take one side of the cut and directly move it to all inferences in the other side where the cut formula occurs as the main formula.  Grade reduction works as you would expect.  Note that since contraction is implicit, the cut rule behaves more like the mix rule in [5].

In contrast to [1], cut-normalization is not strongly normalizing in our calculus.  This is due to the fact that cuts can get "stuck" on certain inferences (equality, reflexivity, induction, definitions, proof links, skolem rules).

In the present implementation, we use a big-step normalization approach: we always directly compute normal forms.  There is a function `evalCut` that takes a cut where both sides are already normalized and then returns a normal form.  Proof substitutions call `evalCut` instead of inserting cut terms.

## Empirical evaluation

We now compare the performance of several cut-elimination procedures on benchmarks used in [2], with a few differences:
 * We measure time in seconds, not in 10<sup>4</sup> milliseconds.
 * We use a consistent color scheme for the different procedures.
 * The proofs generated using `CutIntroduction(LinearEqExampleProof(n))` and `CutIntroduction(FactorialFunctionEqualityExampleProof(n))` have been modified so that they use a binary `E` relation instead of the equality predicate.  This ensures that the CERES methods produce proofs without equational inferences for these proofs.
 * We use a logarithmic scale to compare the runtimes, which differ by several orders of magnitude.

The benchmarks compare the following cut-elimination procedures (the abbreviation is used in the legend of the plots):
 * `LKReductive`: `ReductiveCutElimination`
 * `LKCERES`: `CERES` method that produces proofs in LK
 * `CERESEXP`: `CERES` method that produces expansion proofs, see [1]
 * `BogoElim`: eliminates cuts by forgetting the proof, and generating a new cut-free proof using Escargot.
 * `ExpCutElim`: cut-elimination in expansion proofs, see [6]
 * `LKtNorm`: cut-normalization in LKt
 * `LKtNormA`: cut-normalization in LKt where atomic cuts are not reduced
 * `LKtNormP`: cut-normalization in LKt where propositional cuts are not reduced

### Linear example after cut-introduction

These proofs are generated using `CutIntroduction(LinearExampleProof(n))`.  All of the LKt normalization procedures are faster than the CERES variants by a factor of ~100.  Even BogoElim is faster.  LKt normalization is also faster than expansion proof cut-elimination by a factor of ~10.

ci_linear

### Linear example proof with manual cuts

Cut-introduction often produces unnecessarily complicated lemmas, resulting in irregularity when used in proof sequences.  It is also limited to small proofs.  To produce a more regular sequence and obtain larger proofs, we formalized natural proofs of `P(0), ∀x (P(x) → P(s(x)) ⊢ P(s^{2^n}(x))` using `n` Π₁-cuts.

linear

The results are similar to the proofs obtained with cut-introduction, although we observe new phenomena at both ends of the sequence: for n=0, the proofs consist of a single axiom.  Here, the LKt-based procedures produce a cut-free proof in about 15 nanoseconds.  On the other end, at n>=6, we finally see CERES becoming faster than BogoElim.

### Linear example proof with atomic cuts

To complete the discussion of the linear example proof, we also consider a proof sequence in ACNF.  Interestingly, atomic cut-elimination is surprisingly cheap in this example: the LKt-based normalization only takes 10 microseconds.  On the other hand, the CERES-based methods require as much time as they do for the proofs with Π₁-cuts.  It is unclear where the runtime is spent.

linearacnf

### Square diagonal proof after cut-introduction

These proofs are generated using `CutIntroduction(SquareDiagonalExampleProof(n))`.  LKt normalization upto propositional cuts is an order of magnitude faster than expansion proof cut-elimination, and two orders of magnitude faster than CERES.

ci_sqdiag

### Linear equality example proof after cut-introduction

These proofs are generated using `CutIntroduction(LinearEqExampleProof(n))`.  Note that we replaced the equality predicate by a binary `E` relation to prevent accidental introduction of equational inferences.  Again, LKt normalization upto propositional cuts is 10x faster than expansion proof cut elimination, which is 10x faster than CERES.

ci_lineareq

The astute reader will have noticed the spikes in the runtime of the reductive cut-elimination procedures at n ∈ {7,9,11,13,17}.  These spikes are due to convoluted cut formulas.  For example at n=17, the cut formula is `∀x ((E(f(x), a) → E(f^3(x), a)) ∧ (E(x, a) → E(f(x), a)))` and we use it to prove `E(a,a) → E(f^17(a), a)`--this proof is almost as complicated on the propositional level as the cut-free proof, even though it has a lower quantifier complexity.

## Future work

 * Investigate whether other classical reduction systems such as λμ-calculus or natural deduction (with a suitable normalization relation) can achieve even greater performance.

 * Compare against functional interpretation as in [7].

 * Renaming hypothesis and applying expression substitutions incurs a significant cost in the benchmarks.  An obvious solution is to introduce an explicit substitution inference to implement these operations without the need to traverse the proof term.  Investigate how much this improves performance, or even if.

 * As a cheap optimization, we could grade-reduce blocks of quantifier inferences in a single substitution.

 * We used named variables as a binding strategy since this is traditionally used in GAPT.  Investigate the performance effects of other binding strategies such as DB or LN.

 * Proof assistants such as Coq and Lean also provide functions to normalize proofs.  How does their performance compare to GAPT?  As a realistic upper bound for the performance of cut-normalization, we could also manually convert natural deduction proofs to programs in Scala and measure their runtime.

## References

[1] C. Urban and G. M. Bierman, Strong normalization of cut-elimination in classical logic, Fundamentae Informaticae 44 (2000), 1-34.
[2] A. Leitsch and A. Lolic, Extraction of expansion trees, Journal of Automated Reasoning (2018).
[3] P. Aczel, A general Church-Rosser theorem, Technical Report, University of Manchester (1978).
[4] F. Pfenning, Structural cut elimination, LICS 1995.
[5] G. Takeuti, Proof theory (2013).
[6] S. Hetzl and D. Weller, Expansion trees with cut (2013), [preprint](https://arxiv.org/abs/1308.0428).
[7] P. Gerhardy and U. Kohlenbach, Extracting Herbrand disjunctions by functional interpretation, Archive for Mathematical Logic 44.5 (2005), 633-644.

### DIFF
--- a/core/src/main/scala/at/logic/gapt/expr/Polarity.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/Polarity.scala
@@ -19,25 +19,19 @@ package at.logic.gapt.expr
  * in the antecedent of a clause in a resolution proof has the negative/in-antecedent polarity (as it is in the
  * antecedent of a sequent).  However upon conversion to LK/ET, the polarity switches to positive/in-succedent polarity.
  */
-sealed abstract class Polarity {
-  def inSuc: Boolean
+case class Polarity( inSuc: Boolean ) extends AnyVal {
   def inAnt = !inSuc
 
   def positive = inSuc
   def negative = !positive
 
-  def unary_! : Polarity
+  def unary_! : Polarity = Polarity( !inSuc )
+
+  override def toString = if ( inSuc ) "Positive" else "Negative"
 }
 object Polarity {
-  case object Positive extends Polarity {
-    def inSuc = true
-    def unary_! = Negative
-  }
-  case object Negative extends Polarity {
-    def inSuc = false
-    def unary_! = Positive
-
-  }
+  val Positive = Polarity( true )
+  val Negative = Polarity( false )
 
   val InSuccedent = Positive
   val InAntecedent = Negative
--- a/core/src/main/scala/at/logic/gapt/expr/hol/utils.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/hol/utils.scala
@@ -294,7 +294,8 @@ object instantiate {
     case Ex( v, form ) =>
       val sub = Substitution( v, t )
       sub( form )
-    case _ => throw new Exception( "ERROR: trying to replace variables in a formula without quantifier." )
+    case _ =>
+      throw new Exception( "ERROR: trying to replace variables in a formula without quantifier." )
   }
 
   /**
--- a/core/src/main/scala/at/logic/gapt/expr/substitution.scala
+++ b/core/src/main/scala/at/logic/gapt/expr/substitution.scala
@@ -88,7 +88,7 @@ class Substitution( map: Map[Var, Expr], typeMap: Map[TVar, Ty] = Map() ) extend
       ( this.typeMap.keySet ++ that.typeMap.keySet ).map( v => ( v, this( that( v ) ) ) ) )
 
   def restrict( newDomain: Iterable[Var] ): Substitution =
-    Substitution( newDomain.view.map( v => v -> this( v ) ) )
+    Substitution( newDomain.view.map( v => v -> this( v ) ), typeMap )
 
   def isInjectiveOnDomain: Boolean = isInjective( domain )
   def isInjective( dom: Set[Var] ): Boolean =
--- a/core/src/main/scala/at/logic/gapt/proofs/SequentConnector.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/SequentConnector.scala
@@ -109,6 +109,9 @@ case class SequentConnector( lowerSizes: ( Int, Int ), upperSizes: ( Int, Int ),
     else
       throw new IndexOutOfBoundsException
 
+  def children[T]( upperTs: Sequent[T] ): Sequent[Seq[T]] =
+    inv.parents( upperTs )
+
   /**
    * Given a SequentIndex for the upper sequent, this returns the child of that occurrence in the lower sequent
    * (if there is exactly one).
--- a/core/src/main/scala/at/logic/gapt/proofs/expansion/expansionTrees.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/expansion/expansionTrees.scala
@@ -332,9 +332,9 @@ object ETWeakQuantifierBlock {
         case ETWeakening( _, _ ) =>
       }
 
-    val numberQuants = ( et.polarity, et.shallow ) match {
-      case ( Polarity.InSuccedent, Ex.Block( vs, _ ) )   => vs.size
-      case ( Polarity.InAntecedent, All.Block( vs, _ ) ) => vs.size
+    val numberQuants = et.shallow match {
+      case Ex.Block( vs, _ ) if et.polarity.inSuc  => vs.size
+      case All.Block( vs, _ ) if et.polarity.inAnt => vs.size
     }
 
     walk( et, Seq(), numberQuants )
--- a/core/src/main/scala/at/logic/gapt/proofs/expansion/prenexifyET.scala
+++ b/core/src/main/scala/at/logic/gapt/proofs/expansion/prenexifyET.scala
@@ -7,10 +7,7 @@ import at.logic.gapt.proofs.Sequent
 
 object prenexifyET {
   private def weakQuantifier( polarity: Polarity ) =
-    polarity match {
-      case Positive => Ex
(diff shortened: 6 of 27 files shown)


gebner:
I fixed a performance bug today, and now LKt is (almost) always much faster than expansion proof cut-elimination, which in turn is always faster than CERES.  (Previously the "spikes" were slower than CERES.)

This was due to an oversight: during cut-reduction, it is vitally important to discard inferences that operate on unused auxiliary hypotheses.  In a calculus with explicit weakening, this would correspond to the introduction of weakenings during cut-reduction.

gebner:
As requested by @shetzl, here are the runtimes on the Furstenberg proof:
furstenberg

Interestingly, the runtime is exponential in `n`.

PR #695, opened 2018-03-09
Cut reduction

jannikvierling:


gebner:
There is a massive performance regression in `cutNormal(CutIntroduction(LinearEqExampleProof(7)).get)`, which now takes more than a minute on my machine (10x slower than before).

You can run `cutReductionBenchmark` to see if there are any other obvious performance problems.  It would also be interesting to have a comparison of `cutReductionBenchmark` between the old and new cut-normalization code.

gebner:
BTW, `cutNormalization.scala` has not been scalariform'd.

gebner:
I'll take a closer look at the code next week.

jannikvierling:
I will take a look at the this performance problem.

jannikvierling:
Please try `cutNormal(cleanStructuralRules = true, proof= CutIntroduction(LinearEqExampleProof(7)).get)` on your machine.

jannikvierling:
Now the default behavior is not to clean structural rules.

gebner:
> Please try cutNormal(cleanStructuralRules = true, proof= CutIntroduction(LinearEqExampleProof(7)).get) on your machine.

I'm now seeing just a 10% slowdown on `cutReductionBenchmark`.

jannikvierling:
I think that this slowdown was to be expected, as we are now (in general) reducing more cuts than before. Next week I will run the benchmarks and try to optimize the reduction procedure.

gebner:
@jannikvierling We talked today about whether we should merge this PR in time for the 2.10 release on Monday.  Would it be acceptable for you if we merge this PR *after* the release?

 * I just checked again and there is a 10-50% performance regression on the `cutReductionBenchmark`.  I'd rather not redo the benchmarks and make the LK cut-reduction look even worse than it is now.
 * This leaves very little time for testing.  I'd feel better if we had a few weeks to fix any bugs that crop up (even if that seems unlikely).

jannikvierling:
@gebner Yes, sure. This PR can wait for the 2.11 release.

gebner:
merge?

PR #697, opened 2018-03-12
Spring cleaning: shorten at.logic.gapt to gapt

gebner:
Now that Christmas is nearing its end and the Christmas trees are being taken down, let us celebrate the spring weather with a refreshing cleaning session!

This year I have found something new to remove from GAPT: the `at.logic.` in `at.logic.gapt`.
 * Keeping the domain name in the package name is relatively unidiomatic these days.  Many of our dependencies just have the project name as package name: `scala`, `cats`, `fastparse`, `sourcecode`, `ammonite`, etc.
 * (The most important one:)  I have to type 9 characters less when importing the expr package.
 * It looks more modern.


gebner:
Committed as b5d2b68264b51bb7234139dccd627ff671ce9946.

PR #705, opened 2018-06-13
Tip smt parser

gebner:
Sorry @jannikvierling, I wanted to add some comments so I'm making a PR.


core/src/main/scala/gapt/formats/tip/util/substitute.scala
```diff
@@ -0,0 +1,291 @@
...
+        if ( blacklist.contains( oldName ) ) {
+          nameGenerator.fresh( oldName )
+        } else {
+          oldName
+        }
+      }
+    caseChangeVariableNames( tipSmtCase, oldNames, newNames )
+  }
+
+  type Substitution = Map[TipSmtIdentifier, TipSmtExpression]
```
gebner: NB: this is a path-dependent type (i.e. a different type for each *instance* of `TipSubstitute`.  You may want to put the type alias into the companion object instead.

core/src/main/scala/gapt/formats/tip/util/substitute.scala
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    val substFreeVars: Set[String] =
+      newSubstitution.values.flatMap { freeVariables( problem, _ ) } toSet
+
+    val nameGenerator =
+      new NameGenerator(
+        substFreeVars ++
+          newSubstitution.keys.map { _.name } ++
+          freeVariables( problem, formula ) )
```
gebner: Should we also avoid the function symbols of the problem here?

core/src/main/scala/gapt/formats/tip/parser/tipSmtParser.scala
```diff
@@ -0,0 +1,668 @@
...
+    parse( SExpressionParser( input ) )
+  }
+
+  /**
+   * Parses a TIP problem.
+   *
+   * @param input The input to be parsed.
+   * @return The parsed TIP problem.
+   */
+  def parse( input: String ): TipSmtProblem = {
```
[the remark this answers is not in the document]
jannikvierling: I see. I usually assume that a string represents the content to be parsed and that paths are abstracted by some class because they are system dependent.

core/src/main/scala/gapt/formats/tip/util/substitute.scala
jannikvierling: This is only a temporary solution. I will improve this once I know exactly what this substitution type should look like.

core/src/main/scala/gapt/formats/tip/util/substitute.scala
jannikvierling: Right! I think these symbols should also be avoided, as well as the constructor symbols and the projectors etc. Currently I assume that the input problems are "sufficiently" well-formed to avoid such clashes.

tests/src/test/scala/gapt/formats/tip/TipSmtParserTest.scala
```diff
@@ -3,93 +3,845 @@ package gapt.formats.tip
...
-    val problem = TipSmtParser.parse( ClasspathInputFile( "bin_distrib.smt2" ) )
-    val one = Const( "One", TBase( "Bin" ) )
-    val oneAnd = Const( "OneAnd", TBase( "Bin" ) ->: TBase( "Bin" ) )
-    val instanceSequent = problem.toSequent.map( identity, instantiate( _, Seq( one, one, oneAnd( oneAnd( one ) ) ) ) )
-    Escargot getResolutionProof instanceSequent must beSome
-  }
+  "parsing constant declaration" in {
+
+    "parsing well-formed constant declaration should succeed" in {
+      val constantDeclaration = new SExpressionParser(
```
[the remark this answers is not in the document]
jannikvierling: I will change that. Is there a reason why these implementation dependent methods are visible?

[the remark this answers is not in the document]
gebner: No particular reason.  Feel free to mark the `SExpressionParser` class as private.

PR #710, opened 2018-07-12
Schematic projections and viper input

jannikvierling:


core/src/main/scala/gapt/proofs/Context.scala
```diff
@@ -545,8 +544,13 @@ object Context {
         val Apps( _: Const, ctrArgs ) = lhsArgs( recIdx )
 
         val matchVars = nonRecLhsArgs ++ ctrArgs
-        matchVars.foreach( a => require( a.isInstanceOf[Var] ) )
-        require( matchVars == matchVars.distinct )
+        //Not sure this needs to be here
```
jannikvierling: This check ensures that the left hand side of a definition is of form `f(x_1, ..., x_{recIdx-1}, c(y1, ...,yk), x_{recIdx+1}, ..., x_n)` where `x_1, ..., x_{recIdx-1}, x_{recIdx+1}, x_n, y_1,..., y_k` are pairwise distinct variables. This condition is (implicitly) part of the usual syntax of the primitive recursive functions.

PR #723, opened 2018-10-11
fix ntape printing

quicquid:


quicquid:
Yes, the issue with these particular tests is also that there are no good properties to test apart from that the extraction does not throw errors. Still, the output should have been removed ages ago :-)

PR #729, opened 2018-11-21
Prooftransform

jannikvierling:


core/src/main/scala/gapt/proofs/nd/nd.scala
```diff
@@ -708,11 +708,11 @@ object ImpIntroRule extends ConvenienceConstructor( "ImpIntroRule" ) {
...
   extends BinaryNDProof with CommonRule {
 
-  val negatedFormula = leftPremise( Suc( 0 ) )
-  val formula = rightPremise( Suc( 0 ) )
+  val formulaL = leftPremise( Suc( 0 ) )
+  val formulaR = rightPremise( Suc( 0 ) )
 
-  val mainFormula = if ( negatedFormula == Neg( formula ) ) Bottom() else
-    throw NDRuleCreationException( s"Formula $negatedFormula is not the negation of $formula." )
+  val mainFormula = if ( ( formulaL == Neg( formulaR ) ) || ( formulaR == Neg( formulaL ) ) ) Bottom() else
```
gebner: This seems like a really bad idea.

PR #734, opened 2019-03-28
Superposition with Structural Induction

andreasfrom:
Reference:
[Cruanes S. (2017) Superposition with Structural Induction. In: Dixon C., Finger M. (eds) Frontiers of Combining Systems. FroCoS 2017. Lecture Notes in Computer Science, vol 10483. Springer, Cham](https://doi.org/10.1007/978-3-319-66167-4_10)


shetzl:
Please implement your changes as a viper mode, in parallel to --treegrammar, --analytic, and --portfolio, for example as --spind (superposition and induction). It should be possible to apply your work to a tip file directly from the bash command line.

andreasfrom:
It seems like just reusing the `IndependentInductionAxioms` from `aip` is better than what I cooked up.

andreasfrom:
This can solve `examples/induction/comm2.smt2` now. And associativity of append over three variables but not over one (cf. section 6.2 of Cruanes' paper).

andreasfrom:
I need to remember to run the tests... My little hack to get the clauses out was not good.

core/src/main/scala/gapt/provers/viper/spind/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * Proves the given sequent by using induction.
+   *
+   * @param sequent The sequent to be proven.
+   * @param ctx Defines the constants, types, etc.
+   * @return An inductive proof the sequent is provable with the prover's induction schemes, otherwise None or
+   *         the method does not terminate.
+   */
+  def inductiveLKProof( sequent: Sequent[( String, Formula )] )( implicit ctx: MutableContext ): Option[LKProof] = {
```
jannikvierling: I think it would be easier to generate these induction axioms inside Escargot. Doing this inside Escargot avoids the need to extract the worked off clauses and also gives more control over the saturation process.

andreasfrom:
I'm generalizing passive occurences now, e.g. 2nd, 3rd, 5th and 6th `x` in `x + (x + x) = (x + x) + x` and so can solve that. I'm also doing the thing at the end of section 4.2 where you need to induct on multiple variables if they are primary under the same defined symbol, e.g. `l = take(n, l) ++ drop(n, l)` you need to induct on both `n` and `l` simultaneously. I'm using the `SequentialInductionAxioms` for that, not sure if that's the way to go.

Anyway, I'm maybe missing something because it's not gotten me as far as I'd hoped:
```
mode
analytic_independent    43
analytic_sequential     75
spind                   48
treegrammar             16
```

It's at least better than the independent mode now, but there are six files solved by that which spind fails on.

andreasfrom:
```
mode
analytic_independent    43
analytic_sequential     75
spind                   52
treegrammar             16
```

The five files independent axioms solve:
- isaplanner/prop_26.smt2
 - isaplanner/prop_29.smt2
 - tip2015/int_mul_ident_right.smt2
 - tip2015/list_nat_elem.smt2
 - tip2015/weird_nat_add3_assoc1.smt2

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,475 @@
...
+import gapt.provers.sat.Sat4j
+import gapt.provers.viper.aip.axioms.{ Axiom, SequentialInductionAxioms, StandardInductionAxioms }
+import gapt.provers.viper.grammars.enumerateTerms
+import gapt.utils.NameGenerator
+
+// TODO: sampleTestTerms should probably not be 5 but something dependent on the number of constructors
+case class SpinOptions(
+    performGeneralization: Boolean    = true,
+    sampleTestTerms:       Int        = 5,
+    problem:               TipProblem = null )
```
jannikvierling: Using `Option[TipProblem]` is a better choice than using `null`.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,475 @@
...
+  private def cutAxioms( proof: LKProof, axioms: Seq[Axiom] ): LKProof =
+    axioms.foldRight( proof ) { ( axiom, mainProof ) =>
+      if ( mainProof.conclusion.antecedent contains axiom.formula )
+        CutRule( axiom.proof, mainProof, axiom.formula )
+      else
+        mainProof
+    }
+
+  // Equality is reflexive for all base types
+  def reflRules( ctx: Context ): Set[ReductionRule] = {
```
jannikvierling: The reduction rules provided by the `TipProblem` already contain a rule of the form `x:?a = x -> ⊤`

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,475 @@
...
+
+    val baseTypes = ctx.get[BaseTypes].baseTypes.values
+    val constructors = baseTypes.flatMap( ty => ctx.getConstructors( ty ).getOrElse( List() ) ).toSet
+
+    val same = constructors.map { constr =>
+      val lhsArgs = makeArgs( constr.ty )
+      val rhsArgs = makeArgs( constr.ty )
+
+      val res = lhsArgs.zip( rhsArgs )
+        .foldLeft( Top().asInstanceOf[Formula] ) { case ( acc, ( l, r ) ) => And( acc, Eq( l, r ) ) }
```
jannikvierling: `foldLeft[Formula]( Top() ) { /* ... */ }`

jannikvierling: You could also replace the fold by something like `And( lhsArgs.zip( rhsArgs ) map { case ( l, r ) => Eq( l, r ) } )` which is more readable.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,475 @@
...
+
+      ReductionRule( Eq( Apps( constr, lhsArgs ), Apps( constr, rhsArgs ) ), res )
+    }
+
+    val diff = constructors.flatMap { constr1 =>
```
jannikvierling: Monadic syntax could be more readable in this case
```
for {
  c1 <- constructors
  c2 <- constructors
  if c1 != c2
  if resType( c1.ty ) == resType( c2.ty )
  lhsArgs = makeArgs( c1.ty )
  rhsArgs = makeArgs( c2.ty )
} yield {
 /* ... */
}
```

core/src/main/scala/gapt/expr/BetaReduction.scala
```diff
@@ -62,6 +62,44 @@ object ReductionRule {
   }
 }
 
+// Simple constant simplification and double-negation elimination
```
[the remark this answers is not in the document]
andreasfrom: In `.util` or its own place?

core/src/main/scala/gapt/formats/tip/problem.scala
```diff
@@ -57,9 +76,235 @@ case class TipProblem(
...
+
+  require( !lhs.isInstanceOf[Var], "left hand side must not be a variable" )
+
+  val Apps( lhsHead @ Const( lhsHeadName, _, _ ), lhsArgs ) = lhs
+  val lhsArgsSize: Int = lhsArgs.size
+
+  val allArgs: Set[Int] = lhsArgs.zipWithIndex.map( _._2 ).toSet
+
+  // Argument positions that occur only passively or not at all in expr
+  def passivesIn( expr: Expr, allPositions: Map[Const, Positions] ): Set[Int] = {
```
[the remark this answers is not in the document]
andreasfrom: Sets of `LambdaPosition`? The use of `Set[Int]` follows the methods in `gapt.expr.BetaReduction` and it seems quite natural to me to denote arguments by their index so I can use them directly as indices when matching on `Apps` but I may be missing something.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
andreasfrom: The one I get is special-cased to `i`. That's why I wrote those.
```
scala> problem.reductionRules(8).lhs.toRawString
res7: String = #c(= {i}: i>i>o) #v(x: i) #v(x: i)
```

core/src/main/scala/gapt/expr/BetaReduction.scala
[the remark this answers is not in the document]
jannikvierling: In its own file.

core/src/main/scala/gapt/formats/tip/problem.scala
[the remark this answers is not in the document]
jannikvierling: Yes, I thought that these argument indices are just a special case of Positions ( positions with just one component ). If `Set[Int]` is better suited, then we keep it.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+    def samples( x: Var, f: Expr ): Stream[Expr] = {
+      val constrs = enumerateTerms.forType( x.ty ).filter( _.ty == x.ty ).take( opts.sampleTestTerms )
+      constrs.map( replaceExpr( f, x, _ ) )
+    }
+
+    def go( e: Expr ): Expr =
+      e match {
+        case Ex( x, f ) =>
+          val tests = samples( x, f )
+          tests.foldLeft[Formula]( Bottom() )( ( acc, test ) => Or( acc, go( test ) ) )
```
jannikvierling: You could get rid of the `foldLeft` by something like `Or( tests map { go } )`.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+      rhsArgs = makeArgs( c2.ty )
+    } yield {
+      ReductionRule( Eq( Apps( c1, lhsArgs ), Apps( c2, rhsArgs ) ), Bottom() )
+    }
+
+    same ++ diff
+  }
+
+  // Replace universals and existentials with a fixed number of tests of the formula
+  def unfoldQuantifiers( e: Expr )( implicit ctx: Context ): Expr = {
```
jannikvierling: Do you need `e : Expr`? Wouldn't `e : Formula` suffice?

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+        case App( a, b ) => App( go( a ), go( b ) )
+        case lhs         => lhs
+      }
+
+    go( e )
+  }
+
+  // We need to orient equalities the same way around for the SAT solver to treat them the same
+  def orientEqualities( e: Expr ): Expr = {
+    def go( e: Expr ): Expr =
```
jannikvierling: This inner function is unnecessary.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+        case App( a, b ) => App( go( a ), go( b ) )
+        case lhs         => lhs
+      }
+
+    go( e )
+  }
+
+  // Tests expr by substituting small concrete terms for vars and normalizing the resulting expression.
+  def testFormulaRaw( expr: Expr, vars: List[Var] )( implicit ctx: Context ): Boolean = {
+    def go( e: Expr, subs: List[VarOrConst] ): Seq[Expr] = {
```
jannikvierling: Please add a comment to this method, and consider giving it a more descriptive name.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+
+        case App( a, b ) => App( go( a ), go( b ) )
+        case lhs         => lhs
+      }
+
+    go( e )
+  }
+
+  // Tests expr by substituting small concrete terms for vars and normalizing the resulting expression.
+  def testFormulaRaw( expr: Expr, vars: List[Var] )( implicit ctx: Context ): Boolean = {
```
jannikvierling: This method is rather long and contains many inner functions, maybe we should turn it into an object.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+        case _ if isValid( nf ) || unblock( nf ) => Some( true )
+        case _ if isNormalized( nf )             => Some( false )
+        case _                                   => None
+      }
+
+    val counters = fs.map( normalize ).filterNot { nf =>
+      check( nf ).getOrElse( acceptNotNormalized || constants( nf ).exists( uninterpretedFun( _ )( ctx ) ) )
+    }
+
+    /*
```
jannikvierling: Stray code/comment.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+    }
+
+    val fs = go( expr, vars )
+
+    def normalize( e: Expr ): Expr = orientEqualities( normalizer.normalize( unfoldQuantifiers( e ) ) )
+
+    val origConstants = Context().constants.toSet
+    def isNormalized( e: Expr ): Boolean = constants( e ).forall( c => origConstants.contains( c ) || isConstructor( c ) )
+
+    def isValid( e: Expr ): Boolean = sat.isValid( e.asInstanceOf[Formula] )
```
jannikvierling: I think it would be cleaner to change the argument to `e : Formula` and to do the cast/type checking before invoking this method.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+  // 3: A set of subexpressions that occur in primary positions together directly under the same symbol.
+  //  The sets are transitive, so if a and b occur together and b and c occur together, 3 will contain a set
+  //  containing all of a, b and c.
+  case class Occurences(
+      primary:      Map[Expr, Seq[LambdaPosition]],
+      accumulators: Map[Expr, Seq[LambdaPosition]],
+      passive:      Map[Expr, Seq[LambdaPosition]],
+      underSame:    Set[Set[Expr]] )
+
+  def occurrences( formula: Expr )( implicit ctx: Context ): Occurences = {
```
jannikvierling: This should also be turned into an object.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+    val empty = Seq.empty[( Expr, List[Int] )]
+
+    def newPos( i: Int, size: Int, pos: List[Int] ): List[Int] =
+      2 :: List.fill( size - i - 1 )( 1 ) ++ pos
+
+    var underSame = Set.empty[Set[Expr]]
+
+    def go( expr: Expr, pos: List[Int], inPrimary: Boolean ): ( Seq[( Expr, List[Int] )], Seq[( Expr, List[Int] )], Seq[( Expr, List[Int] )] ) =
```
jannikvierling: Type aliases could make this return type more readable.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+      expr match {
+        case Apps( c @ Const( _, _, _ ), rhsArgs ) =>
```
jannikvierling: When pattern matching and you are just interested in the type, then you could also use `c : Const` instead of `c @ Const(_, _, _)`.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+    val accsPoses = occs.accumulators.filterKeys( asInductiveConst( _ )( ctx ).isDefined )
+
+    accsPoses.foldLeft( f ) {
+      case ( g, ( acc, _ ) ) =>
+        val w = Var( nameGen.fresh( "ind" ), acc.ty )
+        All( w, replaceExpr( g, acc, w ) )
+    }
+  }
+
+  def universalClosureExcept( f: Formula, vars: Set[Var] ): Formula =
```
jannikvierling: This could also be implemented as `All.Block( (variables(f) -- vars).toSeq , f)`.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
```diff
@@ -0,0 +1,486 @@
...
+    ( variables( f ) -- vars ).foldLeft( f ) { case ( g, x ) => All( x, g ) }
```
jannikvierling: Do you really mean `variables`, not `freeVariables`?

core/src/main/scala/gapt/formats/tip/problem.scala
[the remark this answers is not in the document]
andreasfrom: Just checking my understanding: `LambdaPosition` only contains 1 and 2, right? So it would not be positions with one component, but a string of 1's and then a 2.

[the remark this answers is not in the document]
jannikvierling: Yes you are right I forgot about that. The idea about using LambdaPositions here was just to abstract the actual representation of the position.

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
andreasfrom: What do you mean suffice? I thought `Formula` extended `Expr`. And the replacement works on `Expr` anyway so if I want to return a `Formula` again I'll have to cast it.

jannikvierling: It seemed to me more natural to see this operation as an operation on formulas. Clearly it could work on all sorts of terms containing quantifiers. But do we actually unfold quantifiers in such non-formula terms?

core/src/main/scala/gapt/provers/viper/spin/SuperpositionInductionProver.scala
andreasfrom: Definitely `freeVariables`. Thanks!

PR #741, opened 2019-12-10
DLS algorithm for formula equations

jvierling:


jvierling:
@fachammer I created a pull request, so that I can review your code.

jvierling:
@fachammer When committing you can pass multiple options `-m <message>`. The first should give a short description of the commit, the following ones may give a more elaborate explanation of the commit.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+import gapt.expr.formula.hol.HOLPosition
+import gapt.expr.subst._
+import gapt.expr._
+
+object solveFormulaEquation {
+  def apply( formula: Formula ): Try[Formula] = Try( formula match {
+    case Ex( quantifierVariable, innerFormula ) => {
+      // todo: allow multiple argument relations
+      require( quantifierVariable.ty == Ti ->: To, "first existential quantifier variable must be of a relation type" )
+      require( !containsHOQuantifier( innerFormula ), "inner formula must be first order" )
```
jvierling: This check does not agree with its error message. You probably want the condition to be something like `!innerFormula.isInstanceOf[FOLFormula]`.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+import gapt.expr.containedNames
+import gapt.expr.formula.hol.HOLPosition
+import gapt.expr.subst._
+import gapt.expr._
+
+object solveFormulaEquation {
+  def apply( formula: Formula ): Try[Formula] = Try( formula match {
+    case Ex( quantifierVariable, innerFormula ) => {
+      // todo: allow multiple argument relations
+      require( quantifierVariable.ty == Ti ->: To, "first existential quantifier variable must be of a relation type" )
```
jvierling: Checking the preconditions is usually not practical. You may write them down in a comment and rely on the caller to provide appropriate input.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+      val dnf = DNFp( innerFormula )
+
+      val witnessClauses = for {
+        clause <- dnf
+        positiveOccurrences = clause.succedent.filter(
+          atom => atom match {
+            case Atom( variable, _ ) => variable == quantifierVariable
+            case _                   => false
+          } )
+        otherOccurrences = clause.antecedent.map( Neg.apply ) ++ clause.succedent.filterNot( positiveOccurrences.contains )
```
jvierling: The extraction of α and β could be moved to a separate method.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+        witnessComponent = Or( positiveOccurrences.map(
+          atom => atom match {
```
jvierling: The function may be written as `{ case Atom(_, args) => /*...*/ }`

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+            // todo: handle multiple arguments
+            // todo: unique name for unbound variable "x"
+            case Atom( _, args ) => Eq( FOLVar( "x" ), args.head )
+          } ) )
```
jvierling: The computation of a witness component G may also be factored into a separate method.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+      } yield ( witnessComponent, otherOccurrences.map( occurrence => normalize( Substitution( Map( quantifierVariable -> Abs( FOLVar( "x" ), witnessComponent ) ) )( occurrence ) ) ) )
+
+      val result = witnessClauses.tail.foldLeft( ( Imp( And( witnessClauses.head._2 ), witnessClauses.head._1 ), Vector( Neg( And( witnessClauses.head._2 ) ) ) ) )( ( current, next ) => {
```
jvierling: Please try to avoid very long lines.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,53 @@
...
+        ( current, next ) match {
+          case ( ( witness, implicationPrefix ), ( witnessComponent, otherOccurrences ) ) => {
+            ( And( witness, Imp( And( implicationPrefix ++ otherOccurrences ), witnessComponent ) ), implicationPrefix :+ Neg( And( otherOccurrences ) ) )
+          }
+        }
+      } )
```
jvierling: The functions `init`, `last`, and `inits` could be useful for the construction of _S_.

fachammer:
@jvierling thanks a lot for the feedback! Actually, the code was not intended for review yet, since there are quite some rough edges at the moment. But the feedback is very helpful nevertheless since I'm in the process of learning to write idiomatic Scala code. I'll try to incorporate the review when I get to it hopefully later this week.

Furthermore I have some questions regarding the Travis CI checks:
- The [push check](https://travis-ci.org/gapt/gapt/builds/622827092?utm_source=github_status&utm_medium=notification) seems to fail because a failing JDK install script. Do you have an idea what's causing this?
- The [pr check](https://travis-ci.org/gapt/gapt/builds/623364502?utm_source=github_status&utm_medium=notification) is failing because of a failing test which I wrote. However, I cannot reproduce the failing test locally. Seems to be the exact same code. Do you have any idea what could cause the difference between local machine and Travis CI? It might be easier to debug this next week in person though.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
fachammer: Yes, maybe we can even provide an interface that does not allow to pass invalid formulas. For example, by having a datatype that encapsulates the structure of formula equations (first order formula + second order variables). And then maybe provide another method that can extract a formula equation from a Formula. Do you have any thoughts on this?

jvierling:
@fachammer 

- The [push check](https://travis-ci.org/gapt/gapt/builds/622827092?utm_source=github_status&utm_medium=notification) fails because it tries to install JDK 8 which is not supported by Travis anymore.

- The [pr check](https://travis-ci.org/gapt/gapt/builds/623364502?utm_source=github_status&utm_medium=notification) fails because the ordering or the clauses seems to change for some reason (probably the implementation of Set is different depending on the version of Scala that is being used). This should not be a problem as we can order the clauses as we like however the output is then not propositionally equivalent to the expected output. What we actually should require is ∃XF[X] ⇔ F[S].

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
jvierling: Ruling out invalid inputs with static checks is often a good idea. Ideally the function should also be easy to use and its signature should be self explanatory. In this respect I do not see yet how to significantly improve the situation.

fachammer:
@jvierling I think the main part of the DLS algorithm is mostly done, so I'd appreciate some feedback when you have time. 

Some possible next steps in my opinion would be
- considering the negative part of disjuncts to construct a witness. This should make it possible in some cases to also find a witness when using the positive part would require skolemization,
- adding more thorough preprocessing which is stated in the algorithm (renaming variables such that no variable occurs both bound and free and all quantified variables are different and removing redundant quantifiers) which we discussed shortly last time but decided to leave out at the moment, 
- implementing the skolemization/unskolemization part of the algorithm, but as we decided last time we should probably discuss this before and
- adding more unit tests.

If you have any preferences about what I should do next, let me know. Otherwise my next step for now will be to add more tests to cover some corner cases and more complex formula equations.

jvierling:
@fachammer I am currently looking through the code. I think that it would be a good idea to implement the preprocessing steps that produce the variable normal form and remove redundant quantifiers, as these may also be useful for other applications.
Before implementing more complicated things such as skolemization/deskolemization we should clarify whether we are also interested in cases where a witness is not (easily) obtainable but where the quantifier can be eliminated nevertheless.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+    case AndOr( alpha, beta, connective ) =>
+      val movedAlpha = moveExistentialQuantifiersUp( alpha )
+      val movedBeta = moveExistentialQuantifiersUp( beta )
+      if ( movedAlpha != alpha || movedBeta != beta )
+        moveExistentialQuantifiersUp( connective( movedAlpha, movedBeta ) )
+      else
+        formula
+
+    case _ => formula
+  }
```
jvierling: This could be placed in a separate object as this may also be useful elsewhere.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+    formula:      Formula ): Formula = {
+    BetaReduction.betaNormalize( substitution( formula ) )
+  }
+
+  def preprocess(
+    secondOrderVariable: Var,
+    formula:             Formula ): ( List[FOLVar], Set[HOLSequent] ) = {
+    val nnf = toNNF( formula )
+    moveQuantifiers( nnf ) match {
+      case FirstOrderExBlock( variables, innerFormula ) =>
```
jvierling: This case distinction may be replaced by the pattern definition `val FirstOrderExBlock(variables, innerFormula) = moveQuantifiers( nnf )`

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+    moveExistentialQuantifiersUp( moveUniversalQuantifiersDown( formula ) )
+  }
+
+  private def moveExistentialQuantifiersUp( formula: Formula ): Formula = formula match {
+    case Or( Ex( variableAlpha, alpha ), Ex( variableBeta, beta ) ) =>
+      val ( newVariable, substitutedFormulas ) = replaceVariablesWithNewVariable(
+        List(
+          ( variableAlpha, alpha ),
+          ( variableBeta, beta ) ) )
+      Ex( newVariable, moveExistentialQuantifiersUp( Or( substitutedFormulas ) ) )
```
jvierling: The fusion of two quantifiers could be factored into a separate method.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+      else
+        formula
+
+    case _ => formula
+  }
+
+  private def replaceVariablesWithNewVariable[T](
+    variablesInFormulas: Iterable[( Var, Formula )] ): ( Var, Iterable[Formula] ) =
+    variablesInFormulas match {
+      case ( headVariable: Var, _ ) :: _ =>
```
jvierling: This pattern only covers `List`s but not all possible iterables. Running `replaceVariablesWithNewVariable(ArrayBuffer())` will result in a `scala.MatchError`.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+        val blackListVariables = variablesInFormulas.flatMap( x => variables( x._2 ) )
+        val newVariable = rename( headVariable, blackListVariables )
+        val substitute = Substitution( _: Var, newVariable )( _: Formula )
+        ( newVariable, variablesInFormulas.map( x => substitute( x._1, x._2 ) ) )
```
jvierling: Alternatively, we could also first define a substitution which maps every variable of `variablesInFormulas.map(_._1)` to `newVariable` and then apply this substitution to the formulas: `variablesInFormulas.map { case(_, f) => subst(f) }`.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,314 @@
...
+
+    case AndOr( alpha, beta, connective ) =>
+      connective(
+        moveUniversalQuantifiersDown( alpha ),
+        moveUniversalQuantifiersDown( beta ) )
+
+    case _ => formula
+  }
+
+  private def isFreeIn( variable: Var, formula: Formula ): Boolean = {
```
jvierling: For better readability we could use the following implicit class
```
  private implicit class VariableHelper(v: Var) {
    def isFreeIn( f: Formula ): Boolean =
      freeVariables( f ).contains( v )
  }
```
Now instead of calling `isFreeIn(v, f)` we can use `v isFreeIn f`.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
fachammer: do you have a suggestion on where to put this object? maybe into a separate file in the gapt.logic.hol package?

jvierling: The package `gapt.logic.hol` seems to be a good choice.

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,238 @@
...
+   *
+   * The return value is a substitution of the occurring second order variables such that
+   * applying the substitution to φ gives a first order formula which is equivalent to
+   * ?X_1 ... ?X_n φ.
+   *
+   * Does not yet work for formulas where a positive occurrence of the second order variable is
+   * inside the scope of an existential quantifier which is itself inside the scope of an
+   * universal quantifier.
+   */
+  def apply( formula: Formula ): Try[( Substitution, Formula )] = Try( formula match {
```
jvierling: The formula contained in the computed value should also be described (or it should be hidden).

core/src/main/scala/gapt/logic/hol/solveFormulaEquation.scala
```diff
@@ -0,0 +1,251 @@
...
+      val substitutionMap = substitution.map + ( secondOrderVariable -> witness )
+      val updatedSubstitution = Substitution( substitutionMap )
+      val updatedInnerFormula = FirstOrderExBlock( existentialVariables, firstOrderFormula )
+      ( updatedSubstitution, updatedInnerFormula )
+    case _ => ( Substitution(), formula )
+  } )
+
+  private object StrictSecondOrderRelationVariable {
+    def unapply( variable: Var ): Option[( Var, ( Seq[Ty], Ty ) )] = variable match {
+      case Var( _, FunctionType( To, inputTypes @ _ :: _ ) ) if inputTypes.forall( _ == Ti ) =>
```
jvierling: The algorithm should also work if `X` has other argument types than `Ti`.

fachammer:
@jvierling It seems I have been too eager with making the simplification more powerful and now some tests fail. As I see it, most of the tests fail because I remove implications in a formula by using `toNNF`, but the tests expect to have implications still in place after simplify. However, if we don't do that it means we can't so easily simplify formulas like `(P -> Q) ∨ P` (could be simplified to `⊤`) or `(P -> Q) ∧ ¬P` (could be simplified to `¬P` by absorption). Or at least, I don't see how to do this easily without introducing a lot of case distinctions. Do you have an idea on how to do this?

shetzl:
In my opinion, it would be enough to check that the formula output by the DLS implementation is indeed a solution, i.e., that the formula obtained from inserting the output is valid. This can be checked, e.g., by Escargot.isValid.

jvierling:
@fachammer I am currently looking at the problem. The modifications of the simplification procedure you describe seem reasonable, but apparently some parts of our code seem assume that `simplify` preserves in some sense the shape of the formula. I need to look a bit more closely at the code in order to decide how to resolve the problem.

jvierling:
@fachammer I think our best course of action is to create a new simplifier for the DLS algorithm, that implements exactly the simplifications that we need for the FE solver. The existing simplifier was intended to get rid of truth constants and idempotent applications of connectives. Furthermore its current uses suggest, that it was meant to satisfy the postcondition that the simplified formula is propositionally equivalent to the original formula (unfortunately the name `simplify` does not give us much of a hint on that meaning and its documentation is also rather imprecise concerning future extensions). This postcondition is clearly not the valid anymore when doing an NNF transformation. In order to avoid violating this assumption and thus running into compatibility issues, it seems preferable to create a separate simplifier. At a later point in time we might, then factor the simplification code into a more flexible simplification system.

fachammer:
@shetzl @jvierling the latests commits produce significantly smaller witnesses for the example we discussed:
```
∃X_0
  ∃X_1
  ((X_0(x_1) → X_0(x_2) ∧ X_0(s(s(x_2)))) ∧
      (X_1(x_2) → X_1(0) ∧ X_1(s(s(s(s(0)))))) ∧
      P(0) ∧
      (P(x_1) → P(s(x_1))) ∧
      (P(s(x_1)) → P(s(s(x_1)))) →
    P(s(s(s(s(s(s(s(s(0))))))))))
```
As a result substitution we now get:
```
Substitution(
  X_0:i>o -> λx_0_0 (x_1 != x_2 ∨ s(s(x_2)) = x_1 ∨ x_0_0 = x_1) ∧ (x_1 = x_2 ∨ x_0_0 = x_1), 
  X_1:i>o -> λx_1_0 (0 != x_2 ∨ s(s(s(s(0)))) = x_2 ∨ x_1_0 = x_2) ∧ (0 = x_2 ∨ x_1_0 = x_2))
```
Also the computation seems to be faster now.

It seems the issue with this example was that transforming the formula to the form with the smaller formula equation disjuncts produced a lot of disjuncts that didn't even contain the second order variable for which we wanted to have a witness. This produced a big witness due to the witness combination of all the individual witnesses since the witness combination also includes substituting the original first order part. Now, we only combine the individual witnesses for those formula equations that actually contain the second order variable.

jvierling:
@fachammer The pull request is soon ready to be merged. But before that I would like to make some minor changes if that is OK for you.

Among others the changes include:
- I would like to rewrite the section of the user manual a bit, so that fewer technical details (such as e.g. return values, return types, names, etc) actually appear in the text. This should improve the maintainability in case of changes to the methods. Also I would like to make the example slightly more complicated than `?X X(a)`.
- I will probably rename `witnessForSecondOrderQuantifierElimination` to something shorter such as `dls` (or `dls_` because it is a variation of the DLS algorithm). I think that the name `dls` is nice for the API because it explains the underlying principles and limitations of the algorithm.
- Minor refactoring.

fachammer:
@jvierling Yes sure, sounds good to me

core/src/main/scala/gapt/logic/hol/dls.scala
```diff
@@ -0,0 +1,350 @@
...
+    val disjuncts = extractDisjuncts( movedQuantifiersFormula ).filter( _.contains( secondOrderVariable ) )
+
+    val polarizedConjunctsInDisjuncts = disjuncts
+      .map( polarizedConjuncts( _, secondOrderVariable ) )
+    if ( polarizedConjunctsInDisjuncts.exists( _.isEmpty ) )
+      throw new Exception(
+        s"""formula cannot be separated into positive and negative conjuncts of occurrences of $secondOrderVariable
+           |formula: $formulaWithoutRedundantQuantifiers""".stripMargin )
+    else {
+      val ( occurrenceDisjuncts, nonOccurrenceDisjuncts ) = polarizedConjunctsInDisjuncts.map( _.get ).partition( d => d.exists( _.contains( secondOrderVariable ) ) )
```
jvierling: @fachammer As far as I can see `nonOccurrenceDisjuncts` will always be empty because of the filter on line 96. Hence the code below computing the added set could be removed. Is this what you intended?

fachammer: @jvierling Yes, this can be removed now. I just oversaw this when adding the filter.

jvierling: @fachammer :+1:

PR #750, opened 2020-08-18
push to sbt version 1.3.13 and scala 2.13.3

quicquid:


quicquid:
I had trouble getting gapt to run under OpenJDK 14 and ended up at a discussion indicating a scala / sbt version incompatibility. Scala 2.13.3 introduces three kinds of new warnings (that we treat as errors) that affects gapt:

* infix operators with a binary argument are disallowed: this leads to rewrites of e.g. skolemFunction + (x,y) to the uglier skolemFunction.+(x,y) . Should I rename the operators? If yes, which name should I use?
* conversion of Int to Float may lose precision: I've only seen this in prooftool code and at one point in TreeGrammarProver (canSolSize(Int)) that indicates this is also a GUI setting. I consider this as acceptable.
* the `val Bottom() = subProof.conclusion( idx )` pattern matching is supposed to be wrapped into a locally declaration. From the examples I found online, I don't really understand how this applies in this case. Should I keep it like this?

Could you check the changes make sense to you? I'm in particular uncertain if I got the operator priority of * vs + in `proofs/lk/LKProofSubstitutable.scala` right. Thanks for your help!

quicquid:
Merged into develop.

PR #753, opened 2020-11-25
Some tweaks to the tree-grammar based induction prover

gebner:
This is benchmarking code from my thesis that I still had lying around.


jvierling:
@gebner I think you don't need to merge `origin/master`.

gebner:
Ah, that was more out of habit than anything else.  The repo organization has changed quite a bit since I've left.

jvierling:
It is mostly the branching model that has changed. The changes were intended to make the release process more flexible.

PR #766, opened 2023-10-09
Make build compatible with JDK 21

quicquid:
Sbt 1.6.2 and Scala 2.13.8 seem to crash with strange internal exceptions under JDK 21. Bump the versions to current ones that are compatible. Also ThreadDeath / Thread.stop() have been [deprecated](https://docs.oracle.com/javase/8/docs/technotes/guides/concurrency/threadPrimitiveDeprecation.html) for a while and the workaround with the StoppableWithoutDeprecationWarning trait seems to fail. Replace it by the recommended Thread.interrupt() and InterruptedException. The test case for java code succeeds but I have not thested if this also works with external processes like EProver. Also make type annotations of implicit val / var / def explicit to comply with portability warnings regarding Scala 3.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,7 +85,7 @@ lazy val root = project.in( file( "." ) ).
     scalacOptions in ( ScalaUnidoc, unidoc ) ++= Seq(
       "-doc-title", "gapt",
       "-doc-version", version.value,
-      "-doc-source-url", s"https://github.com/gapt/gapt/blob/${"git rev-parse HEAD" !!}/€{FILE_PATH}.scala",
+      "-doc-source-url", s"https://github.com/gapt/gapt/blob/${("git rev-parse HEAD" !!).strip}/€{FILE_PATH}.scala",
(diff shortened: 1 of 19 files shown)


quicquid:
I have tested this only on my own machine, I'd appreciate if it could be verified on a second machine. We might also want to apply this to the master branch.

fachammer:
@quicquid Currently I'm in the process of setting up the build infrastructure again. There seem to be some incompatibilities with the newer scala / sbt and some sbt plugins we use. If you don't mind I'll take this opportunity to fix these by directly pushing to this pull request.

fachammer:
@quicquid The pull request looks good to me. I can confirm the issue when running the gapt shell with jdk 21 without your changes and I can confirm that the issue is gone after your changes. I've also manually tested that a TimeoutException is thrown on the gapt shell when running EProver inside of withTimeout on a simple example with a very short timeout duration. Also, no exception is thrown on longer timeouts.

Is there something else you'd like to verify before putting this into master? Otherwise I'm happy to merge this.

